Issue #96, opened 2019-04-30
Minimum heap size for each benchmark


vhotspur:
In the [last build](https://travis-ci.com/D-iii-S/renaissance-benchmarks/jobs/196534593), `neo4j-analytics` terminated with `java.lang.OutOfMemoryError`. The amount of RAM each Travis job gets is [under 3G](https://docs.travis-ci.com/user/common-build-problems/#my-build-script-is-killed-without-any-error) so that may not be enough.

Thus we might need to exclude some benchmarks from running on Travis (or run them with smaller workloads).

Any thoughts on this?

Issue #66, opened 2019-04-23
All spark-related benchs may fail if ports are used (spark.port.maxRetries not set)


villazon:
Sometimes, if random ports are already used, benchmarks fail (it just checks once that the port is used and aborts)
When looking at the spark console (`http://localhost:4040/environment/`), the spark.port.maxRetries configuration is not set. 
The current implementation is hardcoding the configuration values (e.g. in `log-regression`)
```
def setUpSpark() = {
    val conf = new SparkConf()
      .setAppName("logistic-regression")
      .setMaster(s"local[$THREAD_COUNT]")
      .set("spark.local.dir", tempDirPath.toString)
      .set("spark.sql.warehouse.dir", tempDirPath.resolve("warehouse").toString)
    sc = new SparkContext(conf)
    sc.setLogLevel("ERROR")
  }
```
One possible solution is to add this `spark.port.maxRetries` in the code:
```
def setUpSpark() = {
    val conf = new SparkConf()
      .setAppName("logistic-regression")
      .setMaster(s"local[$THREAD_COUNT]")
      .set("spark.local.dir", tempDirPath.toString)
      .set("spark.port.maxRetries","16")
      .set("spark.sql.warehouse.dir", tempDirPath.resolve("warehouse").toString)
    sc = new SparkContext(conf)
    sc.setLogLevel("ERROR")
  }
```
After adding this line, you will be able to see the `spark.port.maxRetries` correctly set in `http://localhost:4040/environment/`

 Another solution could be to put the values in a configuration file `conf/spark-defaults.conf` as described here https://spark.apache.org/docs/preview/configuration.html

But this needs to the tested, as the configs should be in the renaissance jars

villazon:
According to this https://stackoverflow.com/questions/44914144/error-sparkcontext-error-initializing-sparkcontext-java-net-bindexception-can/44916827 
one solution is to set the the env var `SPARK_LOCAL_IP = 127.0.0.1`
The issue seems to come on hosts with several network interfaces (https://discuss.hail.is/t/how-do-i-fix-java-net-bindexception-cant-assign-requested-address/115) 

On the docs https://spark.apache.org/docs/latest/configuration.html `spark.driver.bindAddress` is the config entry that we could use in `SparkUtil.scala` to set that value.

PR #110, opened 2019-05-03
Add functional testing

farquet:
- Spark benchmark setup and teardown has been moved to `SparkUtil` to avoid code duplication.
- maxRetries increased for port allocation in Spark (relates to #66)
- Add `--functional-test` that runs much smaller workload (relates to #96)
- WIP : Add smaller workloads for all benchmarks (ATM `actors` and `apache-spark` benchmark groups are done)


farquet:
Commit 7ee7375ee77919225e6f09d170f381b55380012b changes `JavaScrabble` to load the resources in a constructor instead of loading them in the static initializer.
The problem with the static initializer is that with the current code, the first iteration will do more work than the subsequent ones, which is conceptually wrong.

farquet:
All benchmarks but 3 have very quick versions now that will be used when `--functional-test` is passed.
Walking through all benchmarks helped find some issues that have been addressed.

Still TODO : 
  - neo4j-analytics : not trivial to create a smaller dataset. (smaller `edges.json` and `vertices`)
  - scala-dotty : requires smaller set of sources to compile. To be discussed.
  - scala-stm-bench7 : not trivial to create a smaller workload.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  def tearDownSparkContext(sc: SparkContext): Unit = {
+    if (sc != null) {
+      sc.stop()
+    }
+  }
+
+  /* For Spark version on renaissance (2.0.0) the Hadoop version is 2.2.0
+   For Windows, the binary zip is not included in the dependencies
```
axel22: Minor: please use the

```
/**
 * Asterisk at every line.
 */
```

style, to be consistent.

farquet:
Execution time without `--functional-test` : 
```
real	5m2.101s
user	24m55.034s
sys	2m2.512s
```

Execution time with `--functional-test` : 
```
real	2m57.831s
user	12m6.246s
sys	0m42.209s
```

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
farquet: Fixed

PR #113, opened 2019-05-06
Use logo as the favicon

farquet:
Also make sure that the favicon displays nicely on all platforms, browsers and devices.

This PR also adds a mention of the PLDI paper on the main page of the website.


farquet:
Added the Renaissance Code of Conduct.

PR #114, opened 2019-05-06
Keep version at one place only

vhotspur:
Currently, there are plenty of spaces where version information is kept. This PR tries to at least keep in sync version printed by harness and version as determined by the main.sbt.

I am not sure this is the cleanest solution but we have to somehow keep the individual places in sync.

Btw, it also means that the released JARs contain invalid version :-(.

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/MetaInfo.java
+++ b/renaissance-core/src/main/java/org/renaissance/MetaInfo.java
@@ -1,0 +1,31 @@
+package org.renaissance;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.jar.Manifest;
+import java.util.jar.Attributes;
+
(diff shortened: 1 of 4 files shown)


renaissance-core/src/main/java/org/renaissance/MetaInfo.java
```diff
@@ -0,0 +1,31 @@
...
+      manifestStream.close();
+
+      if (version == null) {
+        return UNKNOWN_VERSION;
+      }
+      return version;
+    } catch (IOException e) {
+      return UNKNOWN_VERSION;
+    }
+  }
```
axel22: Nice solution (note that one alternative would have been to generate the `version.scala` file from sbt, as a managed source file).

axel22:
LGTM

PR #116, opened 2019-05-06
Website stripe plot

ceresek:
Looks decent to me when run in local Jekyll but somebody should take a look ...


axel22:
It seems that the branch was deleted.

axel22:
The changeset in the first commit of this PR looks good to me, but it seems that the overall diff is empty for some reason.

ceresek:
Sorry, bad revert, please review #117.

PR #117, opened 2019-05-06
Website stripe plot

ceresek:
Sorry, wrong use of revert and merge, should be OK now.

### DIFF
--- a/website/_layouts/mainrenaissance.html
+++ b/website/_layouts/mainrenaissance.html
@@ -141,7 +141,13 @@ layout: base
   <br/>
   <br/>
 
-  <!-- summary of latest benchmark results -->
+  <div id="measurements" class="text-snippet">
+    <img src="https://github.com/renaissance-benchmarks/measurements/raw/master/stripe.png"/>
+    <p>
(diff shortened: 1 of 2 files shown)


website/_layouts/mainrenaissance.html
```diff
@@ -141,7 +141,13 @@
   <br/>
   <br/>
 
-  <!-- summary of latest benchmark results -->
+  <div id="measurements" class="text-snippet">
+    <img src="https://github.com/renaissance-benchmarks/measurements/raw/master/stripe.png"/>
```
axel22: One detail - with the height 500px, the width will be ~1667px, and our current CSS tiling framework is 960px wide.

If we reduced the width to at least 1100px, the figure would spill only a little, while still being readable.

axel22: Example (by editing the Github page directly in the browser):

screenshot-21-54-10

ceresek: Tried on a local copy but width 1100px does not look very readable in my browser (see screenshot), is it possible we have different zoom settings ?

screenshot

ceresek: If we do not want to exceed the 960px I think we should use two rows.

axel22:
Two rows are probably the right solution.

ceresek:
Will do.

PR #121, opened 2019-05-07
Minor website changes

farquet:
- Update CONTRIBUTION.md to remove unicode characters since they do not render correctly on the website
- Update main page title

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -165,10 +165,10 @@ The current members of the committee are:
 
 - Lubomir Bulej, Charles University
 - Gilles Duboscq, Oracle Labs
-- François Farquet, Oracle Labs
+- Francois Farquet, Oracle Labs
 - Vojtech Horky, Charles University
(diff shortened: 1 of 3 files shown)


ceresek:
To render UTF-8 correctly, we need to change loadRemoteContent in utils.js to properly decode data.content. https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings has a solution, essentially we need to use this:

function b64DecodeUnicode(str) {
    // Going backwards: from bytestream, to percent-encoding, to original string.
    return decodeURIComponent(atob(str).split('').map(function(c) {
        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
    }).join(''));
}

farquet:
Nice finding ! I'll give it a try.

lbulej:
Let's not be afraid of UTF-8 in 2019 :-)

Issue #127, opened 2019-05-08
JDK 9+ support


farquet:
We should make sure that the suite works well with the the most recent JDK versions.
And the travis gate should include a JDK11 test and maybe newer if possible.

I currently see that there is an issue building the suite with JDK11 : 
```
[info] Compiling 10 Scala sources to <suite-path>/benchmarks/apache-spark/target/scala-2.11/classes ...
[error] <suite-path>/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala:43:52: value zipWithIndex is not a member of java.util.stream.Stream[String]
[error]       val sublist = for ((line, num) <- text.lines.zipWithIndex if num < MAX_LINE) yield line
[error]                                                    ^
[error] <suite-path>/renaissance/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala:43:72: value < is not a member of Any
[error]       val sublist = for ((line, num) <- text.lines.zipWithIndex if num < MAX_LINE) yield line
[error]                                                                        ^
[error] two errors found
```

villazon:
I pushed PR #128 to fix this issue. 
Solution is inspired from https://github.com/lampepfl/dotty/issues/5463 and https://github.com/Glavo/dotty/commit/59c49fa2d1f427c24d8c6618132bcf038f13d522

farquet:
Thanks ! This fixes the compilation problem, but I see issues during benchmark executions for JDK9 and JDK11. Let’s keep that issue open until we have fixed all of these and added the Travis check on JDK11.

vhotspur:
Should we close it now as #129 was also merged (bc27f1b550884692358f9ae960551c22514f7541)?

axel22:
Sounds good to me to close.
I'd still let @farquet comment because he's aware of some discussion about JDK 12 (?) problems.

farquet:
I am very surprised that it works on Travis, because locally (on MacOS) I hit an error in a static initializer on JDK11 for Spark benchmarks (everything fine on JDK8, JDK9 and JDK10).
What I hit seems to be this Hadoop bug : https://issues.apache.org/jira/browse/HADOOP-14586
Hadoop is a dependency of Spark and since we depend on an old Spark version, it doesn't include a recent enough version of Hadoop with a fix for this.

```
[error] java.lang.ExceptionInInitializerError
[...]

PR #129, opened 2019-05-08
Travis: test on multiple JDKs

vhotspur:
Opening to track progress.


farquet:
Related to #127

vhotspur:
I propose to merge it in the current state where it

* compiles under JDK8, JDK9, JDK11 and OpenJ9
* runs with JDK8, JDK9 and JDK11 (OpenJ9 broken because of #131)

It also caches the build between the stages so the benchmark execution has smaller delay.

Issue #81, opened 2019-04-26
Clean up suite design and API/SPI elements (tentative)


lbulej:
When looking at the changes related to the reworked class loading (#79), I realized that there are some elements of the suite design and the published SPI/API that I believe need a bit of a clean-up. When it comes to public SPI/API, I have this compulsive urge to (try to) do it right before first publishing the suite, but I do realize that I can't currently give it full attention in the remaining time to release due to other duties, so I'm not trying to pick a fight over it :-) 

As far as benchmarks in the repository are concerned, we can deal with any API clean-ups/changes easily (I volunteer), so I'm mainly concerned for SPI/API users who create their own internal benchmarks/plugins/policies. Admittedly, it will take some time before there are any such users, so we might have quite some time here (all the time in the universe, if nobody cares).

Anyway, I would like to know what you think of what I have in mind:

**Separate out API/SPI classes into a package**
It would contain basically four interfaces that define elements used by the harness and the benchmarks. We could actually use the `org.renaissance.core` package for that purpose, because the package would be actually mostly SPI and a bit of API and the core package does not currently contain anything useful (the `Dummy` benchmark should be moved out of there). The package would include:
- `Benchmark` interface, which represents a benchmark to the harness. It would be renamed from `RenaissanceBenchmarkApi` and instead of `runIteration` method it would have an `execute()` or simply `run` so that we avoid the work `iteration`, which we want to avoid talking about (we are trying to consistently use the word repetition.
- `Plugin` interface. It is currently an abstract class, which I would reserve for providing default implementations, but not for an enterface entity. If there is some base code that would make sense to provide to potential plugin developers, we should make such an abstract class. But we should not force anyone to subclass when providing a plugin.
- `Policy` interface. Again, it is currently an abstract class, which forces anyone who wants to provide a policy to subclass, which should be optional. The class currently creates policy factories but I believe that this is the task for the harness.
- `Config` class, possibly modified or replaced. The `Config` object is currently provided to benchmarks, but it really is a harness config -- it contains things such as `readme` or `printList`  flags. I think that possible consumers are at most plugins or policies, but not necessarily benchmarks. Benchmarks should receive a tailored configuration (or context) object, which tells them where their scratch directory is and what objects to use for input/output, possibly even provide some facade for filesystem operations so that benchmarks can request their resources through a central point. In addition to those, I think we may want to be able to pass various benchmark-specific properties (that the harness does not necessarily need to understand) as key-value pairs. 
[...]

PR #130, opened 2019-05-08
Isolate Dummy into separate project and use version/organization from renaissance-core

lbulej:
This came up while trying to untangle things when working on #81. It does two things, actually.
The first is to move the Dummy benchmark into a separate sub-project, so that it can serve its primary purpose, i.e., to test packaging and class loading. This also untangles the Policy from the Dummy so that I can get to eventually killing its current form.

The second thing, that I spent most of the afternoon swearing at SBT (who cares if it's Turing complete, if it cannot do things that make could do bazillion years ago) is to propagate version and organization from somewhere at the top to dependent projects. When looking around for solutions on how to share for example a file that all projects could read a setting value from (no, SBT can't really do that easily, I long for Ant or make) and discovering that SBT does NOT want you to dynamically discover sub-projects and it does NOT want you to have sub-projects with their own .sbt files, I found a solution that seems to work.

Since all projects (including the root project) depend on renaissance-core, I put a single mention of renaissance version into renaissance-core/version.sbt, which is loaded automatically. Then all the projects which depend on core actually use the value from core, using the `(version in renaissanceCore).value` mechanism. And no, this value cannot be assigned to a val, in can only be used in settings context.

Anyway, I realized that I will have to merge clean-ups piece by little piece, otherwise it becomes impossible to keep track of all the changes that fly around in the master (especially when I move files around).


axel22:
> When looking around for solutions on how to share for example a file that all projects could read a setting value from (no, SBT can't really do that easily, I long for Ant or make)

Note that you could read such a value from a joint file in the same way that `scalafmt` does.

> SBT does NOT want you to dynamically discover sub-projects and it does NOT want you to have sub-projects with their own .sbt files, I found a solution that seems to work.

I didn't understand - what do you mean by dynamically, and that subprojects can't have their `.sbt` files?

> Since all projects (including the root project) depend on renaissance-core, I put a single mention of renaissance version into renaissance-core/version.sbt, which is loaded automatically. Then all the projects which depend on core actually use the value from core, using the (version in renaissanceCore).value mechanism. And no, this value cannot be assigned to a val, in can only be used in settings context.

Yes, that's the way to go, IMO.

> And no, this value cannot be assigned to a val, in can only be used in settings context.

The idea behind settings is that you can override them in the sbt shell, so this is why they're bound to the settings context.

However, you could still read the version from a top-level property files (like `scalafmt`), and assign that to a `val`, and then have the setting set to that `val`.

vhotspur:
LGTM.

PR #135, opened 2019-05-14
Use annotations for benchmark metadata (and a few other things)

lbulej:
This is another PR on my way to handle #81. I'm still trying to smooth out the tangle before I can make other changes, so despite the title, this PR includes several things (roughly in order of appearance):

- The root renaissance project is made to depend on renaissance-core. The dependency is there anyway, but in form of class and method names in strings. With this, we can import and use classes from renaissance-core during the build (and in the root `build.sbt`), so that we can get rid of reflection and names in strings.
- Get rid of `MetaInfo` class and `Renaissance-Version` key in `MANIFEST.MF` and instead set `Specification-Title` (`Specification-Version` and `Implementation-Version` are already set by SBT). We can then query the `org.renaissance` package for values of `Specification-Title` and `Implementation-Version` using `Package` methods. I just hope this works despite the OpenJDK [bug](https://bugs.openjdk.java.net/browse/JDK-8201636) we discovered earlier. Edit: @vhotspur checked and it seems to work.
- Add `Benchmark` interface (currently) providing name space for benchmark metadata annotations. This interface WILL also define the methods that benchmarks have to provide and it will be the abstraction on which the harness should depend (not on `RenaissanceBenchmark` which should provide implementation aid).
- Change benchmarks to use the annotations to provide metadata, and change `RenaissanceBenchmark` and `RenaissanceSuite` accordingly.
- A slew of changes to simplify bits of build and simplify things in `RenaissanceSuite`. For example, the build does not generate the `benchmark-group.txt` file, and the harness tries to use `benchmark-details.properties` for most things.


README.md
```diff
@@ -34,7 +34,7 @@ To run a Renaissance benchmark, you need to have a JRE installed.
 This allows you to execute the following `java` command:
 
 ```
-java -jar '<renaissance-home>/target/renaissance-0.1.jar' <benchmarks>
+java -jar '<renaissance-home>/target/renaissance-org.renaissance.jar' <benchmarks>
```
vhotspur: There should be version, right?

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
```diff
@@ -2,22 +2,21 @@ package org.renaissance.actors
...
 class AkkaUct extends RenaissanceBenchmark {
-  override def description(): String =
-    "Runs the Unbalanced Cobwebbed Tree actor workload in Akka."
 
-  override def licenses(): Array[License] = License.create(License.MIT)
+  // TODO: Consolidate benchmark parameters across the suite.
+  //  See: https://github.com/renaissance-benchmarks/renaissance/issues/27
 
-  override def defaultRepetitions = 24
+  private var numIterations: Int = 10
```
vhotspur: Am I missing something if there is 24 in `@Repetitions`?

README.md
lbulej: Ooops. Yes, a typo in a variable.

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
lbulej: > 
> > A slew of changes to simplify bits of build and simplify things in RenaissanceSuite. For example, the build does not generate the benchmark-group.txt file, and the harness tries to use benchmark-details.properties for most things.
> 
> I was wondering whether it would make more sense to have everything in something like `META-INF/RENAISSANCE` in `MANIFEST.MF`-style format?

That's probably doable, especially if we can put all metadata into a single file. Currently there are two.

> 
> > Add Benchmark interface with annotation interfaces for metadata.
> 
> I believe text files next to the benchmarks would work better (e.g. for web-site generation) but definitely this is better than normal methods.

I understand. In the end, I wanted to avoid potential benchmark authors to deal with description files for this kind of information (workload sizes are a different matter). I take it that we can generate any files necessary for website generation during build.

lbulej: > Am I missing something if there is 24 in `@Repetitions`?

numIterations is an internal loop counter, while repetitions is the external one.

vhotspur: > I take it that we can generate any files necessary for website generation during build.

My line of thought was that it would be nice if the website would be more independent of the *compilation* of the suite. That way, the website would not need to be mirrored under the suite repo and Jekyll could get list of benchmarks etc. without need to compile anything.



> > Am I missing something if there is 24 in `@Repetitions`?
> 
> numIterations is an internal loop counter, while repetitions is the external one.

My bad, sorry.

lbulej: > > I take it that we can generate any files necessary for website generation during build.
> 
> My line of thought was that it would be nice if the website would be more independent of the _compilation_ of the suite. That way, the website would not need to be mirrored under the suite repo and Jekyll could get list of benchmarks etc. without need to compile anything.

We could keep such a generated file under version control and push it to (potentially independent) website repo on demand.

After some thought, it seemed better to me if a benchmark author (human) can have the metadata (including descriptions) next to the implementation class (much of it will not change but the descriptions/repetitions even name/group -- they can be set using annotations as well, the names are automatically derived if the annotations are unset).

We can accommodate the website generators (tools) in a different way.

vhotspur: > > > I take it that we can generate any files necessary for website generation during build.
> > 
> > My line of thought was that it would be nice if the website would be more independent of the _compilation_ of the suite. That way, the website would not need to be mirrored under the suite repo and Jekyll could get list of benchmarks etc. without need to compile anything.
> 
> We could keep such a generated file under version control and push it to (potentially independent) website repo on demand.
> 
> After some thought, it seemed better to me if a benchmark author (human) can have the metadata (including descriptions) next to the implementation class (much of it will not change but the descriptions/repetitions even name/group -- they can be set using annotations as well, the names are automatically derived if the annotations are unset).
> 
> We can accommodate the website generators (tools) in a different way.

That makes sense. If we would dump all this information into YAML (or something similar that Jekyll [can read](https://jekyllrb.com/docs/datafiles/)) and version it, it should work too.

lbulej: Sounds like a plan to me and it should be doable from the build system.

lbulej:
> I was wondering whether it would make more sense to have everything in something like `META-INF/RENAISSANCE` in `MANIFEST.MF`-style format?

BTW, is there a Java or Scala API that can produce MANIFEST.MF-formatted files? Something other than the `java.util.jar.Manifest` class, which can read/write a manifest from/to a stream, but does not provide an API to modify a `Manifest` instance. A Scala one would be fine too, as it would only be needed during build.

vhotspur:
> > I was wondering whether it would make more sense to have everything in something like `META-INF/RENAISSANCE` in `MANIFEST.MF`-style format?
> 
> BTW, is there a Java or Scala API that can produce MANIFEST.MF-formatted files? Something other than the `java.util.jar.Manifest` class, which can read/write a manifest from/to a stream, but does not provide an API to modify a `Manifest` instance. A Scala one would be fine too, as it would only be needed during build.

The following works for me even with `java.util.jar.Manifest`

```java
import java.util.jar.*;

public class MakeManifest {
    public static void main(String[] args) throws java.io.IOException {
        Manifest mf = new Manifest();

        Attributes mainAttrs = mf.getMainAttributes();
        mainAttrs.put(Attributes.Name.MANIFEST_VERSION, "1.0");
        mainAttrs.putValue("my-version", "0.0.1");

        Attributes alphaAttrs = new Attributes();
        alphaAttrs.putValue("one", "1");
        mf.getEntries().put("alpha", alphaAttrs);

        Attributes bravoAttrs = new Attributes();
        bravoAttrs.putValue("one", "1");
        bravoAttrs.putValue("two", "2");
        mf.getEntries().put("bravo", bravoAttrs);

        mf.write(System.out);
    }
}
```

as it prints out

```text
Manifest-Version: 1.0
my-version: 0.0.1

Name: bravo
one: 1
two: 2

Name: alpha
one: 1
```

lbulej:
> > > I was wondering whether it would make more sense to have everything in something like `META-INF/RENAISSANCE` in `MANIFEST.MF`-style format?
> > 
> > 
> > BTW, is there a Java or Scala API that can produce MANIFEST.MF-formatted files? Something other than the `java.util.jar.Manifest` class, which can read/write a manifest from/to a stream, but does not provide an API to modify a `Manifest` instance. A Scala one would be fine too, as it would only be needed during build.
> 
> The following works for me even with `java.util.jar.Manifest`
> 

Silly me. I stopped looking one level of abstraction too early. I did not realize that Attributes is a HashMap. OK then, I can go with that.

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
axel22: > > > I take it that we can generate any files necessary for website generation during build.
> > 
> > 
> > My line of thought was that it would be nice if the website would be more independent of the _compilation_ of the suite. That way, the website would not need to be mirrored under the suite repo and Jekyll could get list of benchmarks etc. without need to compile anything.
> 
> We could keep such a generated file under version control and push it to (potentially independent) website repo on demand.
> 
> After some thought, it seemed better to me if a benchmark author (human) can have the metadata (including descriptions) next to the implementation class (much of it will not change but the descriptions/repetitions even name/group may -- they can be set using annotations as well, the names are automatically derived if the annotations are unset).
> 
> We can accommodate the website generators (tools) in a different way.

I like the sentiment that the authors should have all the metadata close to the implementation, because it decreases the likelihood that they miss an update.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
```diff
@@ -2,52 +2,40 @@ package org.renaissance.apache.spark
...
-import org.apache.spark.sql.DataFrame
-import org.apache.spark.sql.SQLContext
-import org.renaissance.Config
-import org.renaissance.License
-import org.renaissance.RenaissanceBenchmark
-
-import scala.util.Random
-
+import org.apache.spark.ml.{Pipeline, PipelineModel, PipelineStage}
+import org.apache.spark.ml.classification.{
```
axel22: I'm personally not a fan of bulk imports, since I think that this should anyway be done by the IDE.
Does our `scalafmt` configuration like this?

build.sbt
```diff
@@ -122,82 +104,45 @@ def jarsAndListGenerator = Def.taskDyn {
...
-
-def createRenaissanceFormatTask = Def.taskDyn {
-  val formatTasks = for (p <- subProjects) yield scalafmt in (p, Compile)
-  val testFormatTasks = for (p <- subProjects) yield scalafmt in (p, Test)
-  val buildFormatTasks = for (p <- subProjects) yield scalafmtSbt in (p, Compile)
-  flattenTasks(formatTasks ++ testFormatTasks ++ buildFormatTasks)
-}
-
-val renaissanceFormatTask = renaissanceFormat := createRenaissanceFormatTask.value
+addCommandAlias("renaissanceFormat", ";scalafmt;scalafmtSbt")
```
axel22: Did you check that this alias runs `scalafmt` in all subprojects, and all configurations (e.g. `test` configuration), and in the `sbt` project files?

build.sbt
```diff
@@ -122,82 +104,45 @@ def jarsAndListGenerator = Def.taskDyn {
...
-
-def createRenaissanceFormatCheckTask = Def.taskDyn {
-  val formatTasks = for (p <- subProjects) yield scalafmtCheck in (p, Compile)
-  val testFormatTasks = for (p <- subProjects) yield scalafmtCheck in (p, Test)
-  val buildFormatTasks = for (p <- subProjects) yield scalafmtSbtCheck in (p, Compile)
-  flattenTasks(formatTasks ++ testFormatTasks ++ buildFormatTasks)
-}
-
-val renaissanceFormatCheckTask = renaissanceFormatCheck := createRenaissanceFormatCheckTask.value
+addCommandAlias("renaissanceFormatCheck", ";scalafmtCheck;scalafmtSbtCheck")
```
axel22: Same here.

build.sbt
```diff
@@ -122,82 +104,45 @@ def jarsAndListGenerator = Def.taskDyn {
...
-        if (!nonGpl || distroLicense == "MIT") {
-          benchGroupContent.append(name).append("=").append(projectShort).append("\n")
-          benchDetails.setProperty("benchmark." + name + ".description", description)
-          benchDetails.setProperty("benchmark." + name + ".repetitions", repetitions.toString)
-          benchDetails.setProperty("benchmark." + name + ".distro", distroLicense.toString)
-          benchDetails.setProperty("benchmark." + name + ".licenses", licenses.toString)
+
+      // Scan project jars for benchmarks and fill the property file.
+      for (benchInfo <- listBenchmarks(project, allJars)) {
+        if (!nonGpl || benchInfo.distro() == License.MIT) {
```
axel22: As a side, at some point we should replace `nonGpl` with `isGpl`, I think. Easier when you don't have to think with double negations.

project/build.sbt
```diff
@@ -0,0 +1,9 @@
+lazy val renaissanceCore = RootProject(uri("../renaissance-core"))
+
+lazy val renaissanceBuild = (project in file("."))
+  .settings(
+    name := "renaissance-build"
+  )
+  .dependsOn(
+    renaissanceCore
+  )
```
axel22: :sunglasses:

project/build.scala
```diff
@@ -0,0 +1,119 @@
...
+        // MIT-compatible, check next entry
+      }
+    }
+
+    License.MIT
+  }
+
+  def distro(): License = distroFromLicenses(licenses())
+
+  def print(): Unit = {
```
axel22: Should we be returning a string instead of printing (i.e. let the caller decide what to do with the value)?

renaissance-core/src/main/java/org/renaissance/Benchmark.java
```diff
@@ -0,0 +1,90 @@
...
+   * class name (without the package prefix). For example, the default name for
+   * a class named {@code LittleBigBench} will be {@code little-big-bench}.
+   */
+  @Documented
+  @Retention(RetentionPolicy.RUNTIME)
+  @interface Name {
+
+    String value();
+
+  }
```
axel22: Should we ever have anything other than the default? IMO, configurability of the name could lead to confusions.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
```diff
@@ -0,0 +1,90 @@
...
+   * Name of the group the benchmark belongs to. Defaults to dash-separated
+   * package names under org.renaissance. For example, the group name for a
+   * benchmark class in the {@code org.renaissance.jdk.concurrent} package
+   * will default to {@code jdk-concurrent}.
+   */
+  @Documented
+  @Retention(RetentionPolicy.RUNTIME)
+  @interface Group {
+
+    String value();
```
axel22: Same here. Seems that the default convention is sufficient, and makes things simpler.

build.sbt
```diff
@@ -35,67 +38,46 @@ def flattenTasks[A](tasks: Seq[Def.Initialize[Task[A]]]): Def.Initialize[Task[Se
...
-            .map(x => x.toString)
-            .mkString(",")
-          val description = clazz.getMethod("description").invoke(instance).toString
-          val reps =
-            Integer.parseInt(clazz.getMethod("defaultRepetitions").invoke(instance).toString)
-          result += ((kebabCase(clazz.getSimpleName), distro, licenses, description, reps))
+          // Print info to see what benchmarks are picked up by the build.
+          val benchClass = clazz.asSubclass(benchBase)
+          val info = new BenchmarkInfo(benchClass)
+          info.print()
```
vhotspur: Is this intentional or a debugging left-over? Of course, not a big issue.

vhotspur:
I really like the British sense of underestimation that is built into SBT when after loading for about two minutes it prints: `[success] Total time: 0 s, completed ...`.

Anyway, I tried to compile it under [1.8.0_152](https://www.oracle.com/technetwork/java/javase/downloads/java-archive-javase8-2177648.html#agreementDivjre-8u152-oth-JPR) as that is the one where the `MANIFEST.MF` [problem](https://bugs.openjdk.java.net/browse/JDK-8201636) occurs and it works there so that should be okay.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
lbulej: What do you mean by bulk? To me bulk would be anything that includes a `*` (in Java). The `{}` form in Scala list exact classes to import in less lines, so I don't really consider it bulk :-)

Scalafmt was fine with that.

build.sbt
lbulej: From what I tested (just running it in sbt shell), it does -- the original method (the one that generated tasks) way actually did not run scalafmt for meta projects.

The various configurations of SBT are still confusing to me (as did not have the need for them), so I will keep trying to simplify the build files. For me SBT is currently a tool where the documentation is basically useless (as if a language theoretician wrote it, enjoying the myriad concepts, but never really explaining how things work for some more complex use cases, such as, apparently, ours), and getting something done requires trial and error and StackOverflow. I hate when I don't know exactly what I'm doing, but I don't have time to study SBT completely.

I spent hours trying to do simple things, but the problem is that when something doesn't work as expected, I don't know why. To me the aggregation of commands should be a basic thing for a multi-project setup, yet if you try to google it, there are no satisfying answers. None of them actually did what we had there originally (which did not work 100% anyway) and mostly suggested creating an alias.

lbulej:
> LGTM
> 
> I left several comments, most notably, we shouldn't leave the names configurable other than through the class name, IMO.

On the contrary, I would use the derived names as defaults, but we should be able to name benchmarks and their groups independently of implementation classes (and projects which capture dependencies). Don't forget that the names and groups will always be subject to approval from reviewers/maintainers, so there is really no danger of them running haywire.

build.sbt
lbulej: Intentional. If logging was conveniently available (it's not, or I can't find it), I would use it. When playing with the build, it was actually a welcome sight to see what benchmarks were picked up and with what metadata. Compared to the mostly useless output of SBT, this at least tells me that build system found the most important things -- the benchmarks ;-)

I'll have a look if logging could be accessed more conveniently from BenchmarkInfo class (instead of build.sbt), so that it can be just logged.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
lbulej: As noted elsewhere, I would leave it there. I would actually prefer those annotations to be set explicitly -- they are logical things, whereas implementation class and subprojects are implementation and build things. 

Consider for example searching through the codebase for "fj-kmeans" -- you will not find it there. Maybe by accident. But if a benchmark has an annotation `@Name("fj-kmeans")` set, it will be spotted immediately. Also consider that me might get to a point where we will have more subprojects to separate dependencies, but the benchmarks in two or more subprojects will be a logical group.

I don't like implicit things, because they require context. I think it a bad practice for implementation things to serve as implicits for logical things that should be decided by authors (subject to approval).

build.sbt
lbulej: Yes, I was peering at the quizzically as well :-)

project/build.scala
lbulej: Preferably yes. I'm not happy with that method there, because it formats things specifically for the build, so I'll possibly move it to build.sbt, where it makes a bit more sense. It put it here initially so that I could get rid of individual statements in one of the loops.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
lbulej: I would contest the `simpler`, see above. It's a different concern. Am I the only one seeing ghosts? :-)

lbulej:
> Modifying the program to also include
> 
> ```java
> bravoAttrs.putValue("three", "Three lines\nwith: new\nlines...\nand colon :");
> ```
> 
> creates an invalid Manifest that throws `java.io.IOException: invalid header field` when reading back from file.
> 
> So it makes sense to use it for version information, list of benchmarks but for their description we should be careful or use some other format (but if we do not want extra libraries in the launcher, the spectrum is limited anyway to Manifests and properties AFAIK).

Perhaps we could encode the line breaks using literal `\n` to make it look to Manifest like one long value. We would have to encode/decode on store/load, but if we could use just a single file and single format for all the metadata, things would be a bit merrier. I wonder what properties do when you store a multiline value in it. Seeing how things are broken, I would almost welcome YAML or even JSON :-)

lbulej:
> Anyway, I tried to compile it under [1.8.0_152](https://www.oracle.com/technetwork/java/javase/downloads/java-archive-javase8-2177648.html#agreementDivjre-8u152-oth-JPR) as that is the one where the `MANIFEST.MF` [problem](https://bugs.openjdk.java.net/browse/JDK-8201636) occurs and it works there so that should be okay.

Thanks for checking!

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
axel22: I mean that one-import-per-line instead of `{...}`. But, I leave it to you.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
axel22: Maybe we should then mandate the `@Name` annotation, *and* mandate that it corresponds to the kebab-case version of the class' name.
I.e. I agree that it's good to avoid implicit-ness, but I would also avoid against too many axes of configuration here.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
axel22: Ok - in that case, I would have a check that the name respects the rule that the class name corresponds to the name, when kebab-cased.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
lbulej: This is what IDEA did when optimizing imports. You are right that with one-import-per-line it's easier to see diffs. Can scalafmt actually fix that?

axel22: Maybe - perhaps @olafurpg could comment?

lbulej:
> > Modifying the program to also include
> > ```java
> > bravoAttrs.putValue("three", "Three lines\nwith: new\nlines...\nand colon :");
> > ```
> > creates an invalid Manifest that throws `java.io.IOException: invalid header field` when reading back from file.
> 
> Perhaps we could encode the line breaks using literal `\n` to make it look to Manifest like one long value. We would have to encode/decode on store/load, but if we could use just a single file and single format for all the metadata, things would be a bit merrier. I wonder what properties do when you store a multiline value in it. Seeing how things are broken, I would almost welcome YAML or even JSON :-)

Storing such value like this:
```
bravoAttrs.putValue("three", value.replace("\n", "\\x0D"));
```
and reading it back like this
```
mf.getAttributes("bravo").getValue("three").replace("\\x0D", "\n")
```
seems to work and the escaping can be probably abstracted away, so I would try to use the manifest format to create `META-INF/RENAISSANCE.MF` (or `META-INF/RENAISSA.NCE` :-)

build.sbt
lbulej: Not really, only if the logger is somehow passed from a task to the class that might need it. The antics involved look like standing on one hand and using the feet to juggle three clay pots filled with nitroglycerine.

project/build.scala
lbulej: Moved, so it's less of an eyesore in the class.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
lbulej: For now, I have got rid of the `{}` imports manually, and will try to keep IDEA configured that way :-)

renaissance-core/src/main/java/org/renaissance/Benchmark.java
lbulej: I used explicit annotations and will add a build check.

renaissance-core/src/main/java/org/renaissance/Benchmark.java
lbulej: As above, I used explicit annotations and will add a build check.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
olafurpg: There's a Scalafmt rewrite to expand grouped imports https://scalameta.org/scalafmt/docs/configuration.html#expandimportselectors More advanced "optimize imports" is out of scope since Scalafmt is only a syntactic tool

Issue #106, opened 2019-05-02
Possible finagle-chirper crash


axel22:
Some users reported seeing this on some machines:

```
Exception in thread "Thread-938" Failure(connection timed out: localhost/127.0.0.1:51834 at remote address: localhost/127.0.0.1:51834. Remote Info: Not Available, flags=0x08) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: localhost/127.0.0.1:51834, Downstream label: :51834, Trace Id: 3b8504004e328e76.3b8504004e328e76<:3b8504004e328e76 with Service -> :51834
Caused by: com.twitter.finagle.ConnectionFailedException: connection timed out: localhost/127.0.0.1:51834 at remote address: localhost/127.0.0.1:51834. Remote Info: Not Available
	at com.twitter.finagle.netty4.ConnectionBuilder$$anon$1.operationComplete(ConnectionBuilder.scala:99)
	at com.twitter.finagle.netty4.ConnectionBuilder$$anon$1.operationComplete(ConnectionBuilder.scala:78)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:511)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:504)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:483)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:424)
	at io.netty.util.concurrent.DefaultPromise.tryFailure(DefaultPromise.java:121)
	at io.netty.channel.epoll.AbstractEpollChannel$AbstractEpollUnsafe$2.run(AbstractEpollChannel.java:570)
	at io.netty.util.concurrent.PromiseTask$RunnableAdapter.call(PromiseTask.java:38)
	at io.netty.util.concurrent.ScheduledFutureTask.run(ScheduledFutureTask.java:127)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
	at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:335)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897)
[...]

PR #136, opened 2019-05-14
Fix finagle-chirper

Fithos:
This PR implements the comments by @vkostyukov in issue #106 (see https://github.com/renaissance-benchmarks/renaissance/issues/106#issuecomment-492048351). 

Since the crash mentioned in #106 never occurred on the machines at my disposal, additional testing would be helpful. In case the crash still occurs, we can leave this PR open until a proper fix is implemented.

### DIFF
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
@@ -322,6 +322,7 @@ class FinagleChirper extends RenaissanceBenchmark {
       for (feed <- feeds) {
         Await.ready(feed.close())
       }
+      Await.ready(master.close())
     }
   }
 
@@ -426,7 +427,7 @@ class FinagleChirper extends RenaissanceBenchmark {
       Await.ready(cache.close())
     }
     Await.ready(master.close())
-    masterService.close()
+    Await.ready(masterService.close())
   }
 
   override def beforeIteration(c: Config): Unit = {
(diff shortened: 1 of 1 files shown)


axel22:
Should this one be getting closed too:

||||||||||||||0x8B0x8C0x8D0x8E0x8F0x900x910x920x930x940x950x960x970x98⇒⇒⇒⇒⇒⇒⇒⇒⇒⇒⇒⇒⇒⇒f32.absf32.negf32.ceilf32.floorf32.truncf32.nearestf32.sqrtf32.addf32.subf32.mulf32.divf32.minf32.maxf32.copysign

axel22:
cc @vkostyukov

Fithos:
> Should this one be getting closed too:
> 
> https://github.com/renaissance-benchmarks/renaissance/pull/136/files#diff-21e7e8759130ae3ea77f1e2657b0d8e1R284

Good point. Fixed.

Fithos:
> Thanks for fixing this @Fithos! This LGTM.

Thanks for your feedback, @vkostyukov!

ceresek:
Testing, should know by tomorrow if it still leaks ...

ceresek:
No crash or apparent leak after 12+ hours of execution on a machine where otherwise we saw a crash after ~3 hours, so looks like this is fixed. Thank you !

@vkostyukov Vladimir, with your Finagle expertise, would you mind becoming a collaborator ? Obviously we could use an occasional ~~kick~~ hint with our Finagle related code.

Fithos:
I think we can close this PR and go ahead with merging. 
How shall we handle bug fixing wrt. releases? I.e., shall we release a v.0.9.1, according to semantic versioning?

ceresek:
Yes, 0.9.1 makes sense. What I'm not sure is if we should base it off 0.9 or current master ... ?

farquet:
IMO, taking master is a reasonable choice since there were other bugfixes merged since the 0.9.0 release (Spark port allocation or preventing extra work in first iteration of Scrabble 7ee7375ee77919225e6f09d170f381b55380012b).
Moreover, keeping a separate release branch would add extra complexity in maintenance and will have to decide what to cherrypick or not and I don't think it's a too risky move to use master.

Btw, regarding result publishing, we could consider showing the release version number somewhere on the home page next to the charts.

axel22:
I agree, we should keep just one dev branch, IMO.

PR #138, opened 2019-05-15
Store JSON and CSV on Ctrl-C (fixes #119)

### LINKED ISSUE #119: Dump partial CSV/JSON output when benchmark interrupted
The CSV/JSON output is only dumped when the benchmark runs to completion. Consider dumping on external interrupt (CTRL-C) so that results collected so far are not lost ? (Maybe also mark the run with some sort of incomplete flag or non zero exit code in JSON.)

vhotspur:
As long as the shutdown hook (from `Runtime` class) is called this should work quite fine.

Added extra field to JSON output to mark VM termination kind.

### DIFF
--- a/renaissance-harness/src/main/scala/org/renaissance/renaissance-suite.scala
+++ b/renaissance-harness/src/main/scala/org/renaissance/renaissance-suite.scala
@@ -15,8 +15,41 @@ import scala.collection._
 import scala.collection.immutable.TreeMap
 
 object RenaissanceSuite {
+
+  class FlushOnShutdownThread(val results: ResultWriter) extends Thread {
+    override def run(): Unit = {
+      results.storeResults(false)
+    }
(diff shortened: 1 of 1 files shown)


ceresek:
When we eventually do validation, shall we terminate on failed validation ? And if so, will this also be called "forced" termination ? Maybe plan for few more exit codes ?

vhotspur:
> When we eventually do validation, shall we terminate on failed validation ?

Probably not because other benchmarks can still execute successfully.

> And if so, will this also be called "forced" termination ? Maybe plan for few more exit codes ?

This is termination of the whole VM, so I do not see other options. When a benchmark fails validation, the results perhaps should not be stored at all? In this direction, I would wait what the new API/SPI offers as I guess it would affect the `ResultObserver` interface too.

PR #147, opened 2019-06-07
Fix parallelism of als and movie-lens to 4 threads per executor

### LINKED ISSUE #145: Contention on Spark benchmarks on machines with lots of CPUs
Currently, our Spark benchmarks set the number of threads per executors to be equal to the number of available CPUs via `Runtime.getRuntime.availableProcessors`.

On machines with lots of CPUs, this value can be high (e.g. 64 or 128) and this has the consequence to create a lot of contention in the benchmark.
The ALS algorithm, used by `als` and `movie-lens` benchmarks, does a lot of data movement and if we have too many threads, the performance will degrade significantly.

See the (rounded) scores for `movie-lens` for different thread configuration in Spark on my 6 cores Macbook Pro. Note that the `Runtime.getRuntime.availableProcessors` returns 12 because of hyperthreading.

| thread count | score (ms) |
|---|---|
| 64 | 370000 |
| 24 | 42000 |
| 12 | 16300 |
| 5 | 9200 |
| 4 | 9400 |
| 3 | 8200 |
| 2 | 8400 |

The workload size **does not** depend on the thread count.

After a bit of research, it appears that we should not set the number of threads to such a high value. 5 is even suggested as an upper bound to maximise hadoop throughput in [1].

> HDFS Throughput: HDFS client has trouble with tons of concurrent threads. It was observed that HDFS achieves full write throughput with ~5 tasks per executor. So it’s good to keep the number of cores per executor below that number.

I would suggest that we fix the number of threads per executor to 3, 4 or 5 to still exercise parallelism as much as we can but without creating contention for Spark.

@ceresek, since you made some measurements, did you notice such behavior on big machines ?

[1] https://spoddutur.github.io/spark-notes/distribution_of_executors_cores_and_memory_for_spark_application.html

[2] https://db-blog.web.cern.ch/blog/luca-canali/2017-09-performance-analysis-cpu-intensive-workload-apache-spark

farquet:
Fixes #145

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
@@ -27,6 +27,8 @@ class Als extends RenaissanceBenchmark with SparkUtil {
 
   var numRatings = 20000
 
+  val THREAD_COUNT = 4
+
   // TODO: Unify handling of scratch directories throughout the suite.
   //  See: https://github.com/renaissance-benchmarks/renaissance/issues/13
 
@@ -68,7 +70,7 @@ class Als extends RenaissanceBenchmark with SparkUtil {
 
   override def setUpBeforeAll(c: Config): Unit = {
     tempDirPath = RenaissanceBenchmark.generateTempDir("als")
-    sc = setUpSparkContext(tempDirPath)
+    sc = setUpSparkContext(tempDirPath, THREAD_COUNT)
     if (c.functionalTest) {
       numRatings = 500
     }
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
@@ -27,8 +27,6 @@ class Als extends RenaissanceBenchmark with SparkUtil {
 
   var numRatings = 20000
 
-  val THREAD_COUNT = Runtime.getRuntime.availableProcessors
-
   // TODO: Unify handling of scratch directories throughout the suite.
   //  See: https://github.com/renaissance-benchmarks/renaissance/issues/13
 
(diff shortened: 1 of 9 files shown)


axel22:
I'm not sure we should change the parallelism of all the benchmarks. We should have a meeting to discuss this, and we should investigate the scalability of all the benchmarks further.

lbulej:
> I think we should only change als and movie-lens now, and investigate the other benchmarks separately.

I support Alex's view. Such a change should be based on some evidence that everyone can review. I think it should be discussed in a separate issue that we can refer to, and we should keep a separate changelog file in the project that concerns such changes in benchmarks.

farquet:
I'm happy to get more precise numbers for all Spark benchmarks, but independently of what the actual numbers are, **it is a bug** to set the number of threads to something so high and does not represent what an actual Spark workload looks like.

Indeed, a Spark app writer would scale the app by increasing the number of physical nodes and/or the number of executors per node, but won't go beyond 5 for the number of threads per executor because it doesn't give you more throughput and creates contention [1]. So what we are doing here is clearly wrong.

The solution proposed in that PR keeps a single executor and removes this bug. I would argue against the fact that using all cores as spark threads makes it a parallel benchmark, it only makes it a buggy benchmark since there is just a single executor anyway.

Another reasonable approach that would really make those benchmarks parallel using all the available cores would be to increase the number of executors (spark.executor.instances) and give each executors 2 to 4 threads.

Basically, the best way to maximize resource usage and throughput on a single Spark node would be to do something like : `NUM_THREADS * NUM_EXECUTORS = NUM_CORES - 1`.

NUM_THREADS could be different from a benchmark to another depending on how much data movement there is. So this could be fine tuned per benchmark and NUM_EXECUTORS would simply be computed based on NUM_THREADS and NUM_CORES. Maybe a max value could also be set.

I gave a quick try of that approach and I can get 9% better throughput for page-rank for instance if I fix 4 threads and 2 executors on my 6 cores macbook. However, movie-lens becomes much jumpier, probably because of data movement and maybe more GC. Nevertheless, some iterations give a better score than the best score achieved with the old setup.

Anyway, I don't want to dive too deep into this analysis now, but I think it is the way to go. My suggestion here would be to :

Merge that PR as a bugfix to make the suite usable and correct on big machines
Do proper parallelism of Spark benchmarks in a subsequent PR. That would imply a deeper analysis of each better to find the good tradeoffs for each workload.
[1] https://stackoverflow.com/questions/24622108/apache-spark-the-number-of-cores-vs-the-number-of-executors

farquet:
As discussed, I updated the PR to only touch `als` and `movie-lens` parallelism.

PR #154, opened 2019-06-14
Add basic JMH support

axel22:
This commit adds basic support for running the benchmarks from JMH.
I introduced a new project that depends on the main `renaissance` project, and uses it to list all the benchmarks and generate JMH wrapper classes. It then produces a fat JAR similar to the regular Renaissance JAR, but which includes the generated classes and the JMH's generated boilerplate code and resources.


build.sbt
```diff
@@ -243,3 +249,134 @@ lazy val renaissance: Project = {
...
+      case PathList("scala", xs @ _*) =>
+        MergeStrategy.discard
+      case x =>
+        val oldStrategy = (assemblyMergeStrategy in assembly).value
+        oldStrategy(x)
+    }
+  )
+  .dependsOn(
+    renaissance
+  )
```
lbulej: Do you think that these chunks of JMH-specific code live in a separate `.scala` file (say `jmh.scala`) in the project directory?

axel22: That's a good idea - I will move them.

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
```diff
@@ -38,6 +38,7 @@ class AkkaUct extends RenaissanceBenchmark {
   }
 
   protected override def runIteration(config: Config): Unit = {
+    //sys.error(" --> " + Predef.getClass.getClassLoader + " --> " + this.getClass.getClassLoader)
```
lbulej: Not needed anymore?

axel22: Good catch - removed!

build.sbt
axel22: Moved JMH-specific functionality to `jmh.scala`.

lbulej:
The wrapper classes will need to change when the benchmark API changes, so I would sleep better if you also added a test to Travis to test that the JMH version works :-) Other than that, LGTM.

axel22:
Sounds good, let me see if my Travis-fu is adequate :)

axel22:
Seems to work: https://travis-ci.org/renaissance-benchmarks/renaissance/jobs/545994131

project/build.scala
```diff
@@ -110,4 +114,59 @@ class BenchmarkInfo(val benchClass: Class[_ <: RenaissanceBenchmark]) {
...
+
+          val isEligible =
+            !excludePattern.matcher(clazz.getPackage.getName).matches() &&
+              benchBase.isAssignableFrom(clazz)
+          if (isEligible) {
+            // Print info to see what benchmarks are picked up by the build.
+            val benchClass = clazz.asSubclass(benchBase)
+            val info = new BenchmarkInfo(benchClass)
+            if (logger.nonEmpty) logBenchmark(info, logger.get)
+
```
davleopo: empty newline?

renaissance-core/src/main/java/org/renaissance/util/BenchmarkLoader.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public RenaissanceBenchmark loadBenchmark(String name) {
+        try {
+            final Info bench = benchmarksByName.get(name);
+            final ClassLoader loader = ModuleLoader.getForGroup(bench.group);
+            final Class<?> benchClass = loader.loadClass(bench.className);
+            final Object result = benchClass.getDeclaredConstructor().newInstance();
+
+            // Make current thread as independent of the harness as possible
```
davleopo: missing "."

project/build.scala
axel22: Removed.

renaissance-core/src/main/java/org/renaissance/util/BenchmarkLoader.java
axel22: Fixed.

PR #156, opened 2019-06-17
finagle-http bugfixes (fixes #148, fixes #149)

### LINKED ISSUE #148: finagle-http not sending enough requests
Looking at the following code in `benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala` it seems to me that `val response ...` should be moved to the inner loop as in the current setup we sent one request per client and repeatedly read its result. But the naming suggests that there should be `NUM_REQUESTS` requests sent.

Is this a bug or intentional behaviour?

```scala
override def run(): Unit = {
  val client: Service[http.Request, http.Response] =
    com.twitter.finagle.Http.newService(s"localhost:$port")
  val request = http.Request(http.Method.Get, "/json")
  request.host = s"localhost:$port"
  val response: Future[http.Response] = client(request) // <============
  for (i <- 0 until NUM_REQUESTS) {
    Await.result(response.onSuccess { rep: http.Response =>
      totalLength += rep.content.length
    })
  }
  client.close()
}
```

### LINKED ISSUE #149: finagle-http not parallel enough?
Looking at the benchmark in `benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala` I wonder whether the `clientThread.join()` should not be in a separate loop as outlined below? That way the benchmark would spawn multiple (parallel) clients, let them work and wait for them in the end. Or is this intentional?

```scala
val listOfThreads = ...
for (i <- 0 until NUM_CLIENTS) {
  val clientThread = new Thread {
    ...
  clientThread.start()
  listOfThreads.append(clientThread)
}
for (i <- listOfThreads) {
  i.join()
}
```

vhotspur:
finagle-http is now sending requests in parallel. Also, the measured part of the benchmark now excludes starting of the threads (there is a barrier to start all the work at once).

Note that the parameters (number of clients and number of requests) were changed to roughly correspond to the previous performance (same order of magnitude).

Also note that this commit changes the performance of the finagle-http benchmark drastically.

This closes #148 and #149 but there still seems to be some other bug as sometimes exactly one request is lost (see Travis links in #151).

### DIFF
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
@@ -132,6 +132,7 @@ class FinagleHttp extends RenaissanceBenchmark {
     threadBarrier = new CountDownLatch(NUM_CLIENTS + 1)
     for (i <- 0 until NUM_CLIENTS) {
       threads(i) = new WorkerThread(port, threadBarrier, NUM_REQUESTS)
+      threads(i).setName(s"finagle-http-worker-$i")
       threads(i).start()
     }
   }
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
@@ -46,7 +46,9 @@ class FinagleHttp extends RenaissanceBenchmark {
         val request = http.Request(http.Method.Get, "/json")
         request.host = s"localhost:$port"
         val response: Future[http.Response] = client(request)
-        Await.result(response.onSuccess { rep: http.Response =>
+        // Need to use map() instead of onSuccess() as we actually need to
+        // wait for the side-effect, not the original response
+        Await.result(response.map { rep: http.Response =>
           totalContentLength += rep.content.length
         })
       }
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
@@ -2,6 +2,7 @@ package org.renaissance.twitter.finagle
 
 import java.net.InetSocketAddress
 import java.util.Date
+import java.util.concurrent.CountDownLatch
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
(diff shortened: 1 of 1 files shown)


benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
```diff
@@ -31,16 +32,38 @@ import org.renaissance.SimpleResult
...
 
   /** Number of requests sent during the execution of the benchmark.
    */
-  var NUM_REQUESTS = 2000000
+  var NUM_REQUESTS = 12000
 
   /** Number of clients that are simultaneously sending the requests.
    */
-  var NUM_CLIENTS = 20
+  var NUM_CLIENTS = 8
```
farquet: Note that we have a consistency problem here. Here, the number of clients is just a constant, while in `finagle-chirper` the benchmark creates as many parallel clients as the number of available threads : 
https://github.com/renaissance-benchmarks/renaissance/blob/master/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala#L357

We definitely need some reasonable policy about parallelism to we could stick to...

vhotspur: IMHO the approach in `finagle-chirper` is generally better. However, it would mean that all CPUs would be occupied by clients thus the CPUs would be shared with the server threads. Perhaps splitting the threads more evenly would make more sense?

vhotspur:
> Could you share the performance, before and after that change, on a few VMs to get a feeling of how different the benchmark is from the previous version ?

Yes, I will run it on something more powerful than my desktop and share the numbers once they finish.

However the numbers are incomparable because the benchmark is now executing a different workload. But I believe the updated version is a bit more realistic and better corresponds with the purpose of the benchmark.

benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleHttp.scala
axel22: Yes, I think so too. Perhaps the benchmark should determine the parallelism level first:

```
val parallelism = Runtime.getRuntime.availableProcessors
```

And then split those among the agents of the benchmark. I'm wondering where the server thread count is set.

farquet: Could we get this as part of this PR ? And maybe align `finagle-chirper` with the decision we will take here.

vhotspur:
I hope I fixed #161 too. I am now re-running the measurements for comparison after this fix.

vhotspur:
Here are the performance comparison results. Executed on 24 CPU machine with OpenJDK 11 with `-Xms8G -Xmx8G` and otherwise default settings.

Each boxplot is one JVM restart (displaying last 10 iterations out of 40 total), colors distinguish restarts (the ordering of the boxplots should be considered random). Note that the current (`master` branch) version jumps a bit more than branch from this PR (both variability inside restart and between restarts).

finagle-http-compare-master-vs-pr156


Any objections to merging this in its current state?

PR #160, opened 2019-06-18
Add JMH default values for iteration count that correspond more closely to Renaissance defaults.

axel22:
cc @tkrodriguez @dougxc

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -250,21 +250,22 @@ lazy val renaissance: Project = {
   allProjects.foldLeft(p)(_ aggregate _)
 }
 
-def generateJmhWrapperBenchmarkClass(
-  name: String,
-  packageName: String,
-  outputDir: File
-): File = {
+def generateJmhWrapperBenchmarkClass(info: BenchmarkInfo, outputDir: File): File = {
+  val packageName = info.benchClass.getPackage.getName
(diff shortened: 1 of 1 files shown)


project/jmh.scala
```diff
@@ -6,21 +6,22 @@ import sbt.util.Logger
...
      import java.util.concurrent.TimeUnit;
+     import static java.util.concurrent.TimeUnit.MILLISECONDS;
      import org.openjdk.jmh.annotations.*;
      import org.renaissance.RenaissanceBenchmark;
      import org.renaissance.JmhRenaissanceBenchmark;
 
      @State(Scope.Benchmark)
      @OutputTimeUnit(TimeUnit.MILLISECONDS)
+     @Warmup(iterations = ${info.repetitions}, time = 1, timeUnit = MILLISECONDS)
+     @Measurement(iterations = ${info.repetitions / 4 + 1}, time = 1, timeUnit = MILLISECONDS)
```
farquet: How did you choose this value ?

axel22: In our infrastructure, we take about 20-40% of the last iterations to compute the mean value, which seems to give reasonably stable measurements.
The `+ 1` is added to circumvent situations in which `repetitions < 4`.

Issue #46, opened 2019-04-17
Add automation for the website publishing


axel22:
Since we'll be publishing from the `website` directory, we will need a job that invokes `jekyll` and copies the files to `renaissance-benchmarks.github.io`.

This is not critical for the first release, since we can just manually copy the files, but we should eventually automate it.

PR #174, opened 2019-07-03
Release v0.10.0

farquet:
Blog announcement and version bump


axel22:
Some other resources hardcode the `0.9.0` version, currently - for example, the Travis CI definition.
I suggest that you do a grep to find all such occurrences, and update them.

farquet:
> Some other resources hardcode the 0.9.0 version, currently - for example, the Travis CI definition.
I suggest that you do a grep to find all such occurrences, and update them.

Fixed.

farquet:
Interesting. Updating the readme from within sbt with `runMain org.renaissance.Launcher --readme` doesn't provide the same output as using the bundled jar : 
```
diff --git a/README.md b/README.md
index 3aad9d2..038b28a 100644
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ For example, you can specify `scala-kmeans` as the benchmark.
 The following is a complete list of command-line options.

-renaissance-core, version 0.10.0
+Renaissance Benchmark Suite, version 0.10.0
 Usage: renaissance [options] [benchmark-specification]

   -h, --help               Prints this usage text.
```

.travis.yml
```diff
@@ -15,20 +15,20 @@ refs:
       - 'mkdir -p "$HOME/.prebuilt"'
       - 'ls "$HOME/.prebuilt"'
       - 'rm -rf "$HOME/.prebuilt"/*.jar'
-      - 'cp target/renaissance-gpl-0.9.0.jar "$HOME/.prebuilt/"`git rev-parse HEAD`.jar'
```
axel22: Opt: perhaps there is a way to add an env variable to this file, to have a single version declaration.
If you do, I would suggest adding a comment to `version.sbt`, with a reminder that `.travis.yml` should also be changed when changing `version.sbt`.

axel22:
> Interesting. Updating the readme from within sbt with `runMain org.renaissance.Launcher --readme` doesn't provide the same output as using the fatjar :
> 

I *think* that you should change this line to change the title:

```
  val renaissanceTitle = classOf[RenaissanceBenchmark].getPackage.getSpecificationTitle
```

.travis.yml
farquet: That's a good idea. I'll do it.
Let's try to make the release process as easy as possible. Btw, coffee is on me for the one who will take care of #46 ;)

PR #178, opened 2019-07-10
Store benchmark failure into JSON (issue #162)

### LINKED ISSUE #162: Validation failure not stored in JSON/CSV
When a benchmark (validation) fails, the JSON output is shorter (than in successful run) but the failure is not logged there. One has to inspect stdout for that information.

This information should be part of the JSON somehow (we already have `environment.vm.termination` field there).

Storing this inside CSV could be tricky, though.

vhotspur:
This is an attempt to fix #162 (JSON and CSV do not report in any way whether the benchmark failed).

In this PR the information about benchmark outcome is added to JSON. I have not added it to CSV as I see no reasonable way to store it there.

For the purposes of testing, this PR also extends the dummy benchmarks with two failing ones.

The downside is that the code for the output writers is growing in complexity. Hopefully with #81 this functionality could be moved to separate plugins.

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
+++ b/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
@@ -109,6 +109,7 @@ public abstract class RenaissanceBenchmark {
         for (ResultObserver observer : config.resultObservers()) {
           observer.onFailure(name());
         }
+        return t;
       }
     }
   }
--- a/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
+++ b/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
@@ -96,6 +96,9 @@ public abstract class RenaissanceBenchmark {
       }
       return null;
     } catch (Throwable t) {
+      for (ResultObserver observer : config.resultObservers()) {
+        observer.onFailure(name());
+      }
       return t;
     } finally {
       try {
--- a/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
+++ b/renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
@@ -106,6 +106,9 @@ public abstract class RenaissanceBenchmark {
       } catch (Throwable t) {
         System.err.println("Error during tear-down: " + t.getMessage());
         t.printStackTrace();
+        for (ResultObserver observer : config.resultObservers()) {
+          observer.onFailure(name());
+        }
       }
     }
   }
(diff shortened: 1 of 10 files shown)


renaissance-core/src/main/java/org/renaissance/ResultObserver.java
```diff
@@ -11,5 +11,13 @@
...
   public void onNewResult(String benchmark, String metric, long value);
+
+  /** Callback when benchmark fails.
+   *
+   * Serves to e.g. invalidate previous results or marking the run as failed.
+   *
+   * @param benchmark Name of the benchmark.
+   */
+  public void onFailure(String benchmark);
   public void onExit();
```
axel22: We should add an empty line.

renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
```diff
@@ -96,6 +96,9 @@ public final Throwable runBenchmark(Config config) {
       }
       return null;
     } catch (Throwable t) {
+      for (ResultObserver observer : config.resultObservers()) {
+        observer.onFailure(name());
+      }
```
axel22: What would be the consequences of placing this callback in the `finally` block after the `tearDownAfterAll` call?

renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
```diff
@@ -96,13 +96,19 @@ public final Throwable runBenchmark(Config config) {
...
       return t;
     } finally {
       try {
         tearDownAfterAll(config);
       } catch (Throwable t) {
         System.err.println("Error during tear-down: " + t.getMessage());
         t.printStackTrace();
+        for (ResultObserver observer : config.resultObservers()) {
+          observer.onFailure(name());
```
axel22: Another thing - should you be passing the exception object to the observer too?

axel22: Also, should `t` be returned here?

renaissance-core/src/main/java/org/renaissance/ResultObserver.java
vhotspur: Done, thanks for catching this.

renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
vhotspur: It should be there too :-). Fixed.

renaissance-core/src/main/java/org/renaissance/RenaissanceBenchmark.java
vhotspur: > Also, should t be returned here?

I am not sure why we do not return `t` here but it definitely makes sense. Added.

> Another thing - should you be passing the exception object to the observer 

Maybe, it is not needed now.

PR #179, opened 2019-07-10
Travis: check *.java are in pure ASCII

### LINKED ISSUE #43: Prevent non ASCII characters in Java source files
Non english users on Mac OS will hit 
```
[error] <path>/renaissance-benchmarks/benchmarks/rx/src/main/java/org/renaissance/rx/RxScrabbleImplementation.java:2:1: unmappable character for encoding ASCII
```
if they run `sbt assembly` since a non ASCII character appears in the source code because of locale environment variables set automatically by the OS.

We could add a travis check that prevents non ascii characters in java files (scala is not a problem).
The goal being to make the process as smooth as possible for new cloners of the repo.

vhotspur:
This should fix #43 where non-ASCII characters caused compilation problems on Mac.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,11 @@ jobs:
         - ./tools/sbt/bin/sbt assembly
         - java -jar ./target/renaissance-gpl-$RENAISSANCE_VERSION.jar --readme && git diff --exit-code -- README.md CONTRIBUTION.md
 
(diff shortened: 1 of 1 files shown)


vhotspur:
Hmmm :-(. Not sure why Travis CI was not triggered on this PR.

Issue #27, opened 2019-04-12
Decide on benchmark-specific parametrization


axel22:
Should the benchmarks have configurable sizes and parallelism levels, or other parameters?
If so, how should a benchmark implementation inform the harness about its available parameters, and how should these be passed from the harness to the benchmark?

vhotspur:
Implemented via #191.

PR #182, opened 2019-07-15
WiP: benchmark parameters (configurations)

vhotspur:
Start work on benchmark parameters (#27). It is now possible to define benchmark configurations via

```java
@Configuration(name="default", values={"param=42", "threads=1"})
@Configuration(name="large", values={"param=42", "threads=cpu.count"})
```

added to the benchmark class and then (from inside the benchmark) query them via

```java
config.parameter("threads")
```

to alter benchmark workload (user specifies the configuration via `--configuration`).

Currently, only integers are supported as parameter values and use of `Config` to pass the parameters is a temporary workaround as they belong more to some context object (see #81).

Added benchmark `dummy-params` as a demonstrator.

Let's take this as a starting point for discussion how it should look like. The implementation is minimal but it allows at least some experimenting.


lbulej:
Integrated (in slightly modified form) into the api-cleanup branch, including benchmark modifications.

lbulej:
Superseded by #191

PR #183, opened 2019-07-15
dotty: rework classpath handling (fixes #176)

### LINKED ISSUE #176: dotty benchmark fails on Windows
Running `dotty` benchmark on Windows results in the following failure (see [also here](https://ci.appveyor.com/project/vhotspur/renaissance/builds/25722034#L694) a build from AppVeyor (PR #139)):

```
java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:/C:/Users/user/renaissance/./tmp-jars-76485191019216267
05/cp-scala-dotty_2.12-0.9.0.jar5631486227090191329.jar                                                                         
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)                                                   
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)                                                       
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)                                                        
        at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)                                                                    
        at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)                                                     
        at java.nio.file.Paths.get(Paths.java:84)                                                                               
        at dotty.tools.io.Directory$.apply(Directory.scala:30)  
        ...
```

For context, we run the compiler with the following arguments:

```scala
    val args = Seq[String](
      DOTTY_ARG_CLASS_PATH,
      Thread.currentThread.getContextClassLoader
        .asInstanceOf[URLClassLoader]
        .getURLs
        .mkString(File.pathSeparator),
      DOTTY_ARG_TYPE_CONVERSION,
      DOTTY_ARG_CLASS_FILE_DESTINATION,
      outputPath.toString
    )
```

The culprit seems to be that in the end we pass list of URLs that are preceded by `file:/` that the Windows part of FS abstraction is unable to handle. Note that the harness installed this context class loader so we know that the classpath points always to a list of JARs stored on disk.

It seems that the following fixes it (it still runs on Linux and it also runs on Windows). But that is a rather fragile solution. I do not know Dotty at all so I am wondering why it cannot use the existing classpath or whether it would be possible to give it some classloader directly instead of this remarshalling.

Any opinions what to do with this (especially whether we should merge such workaround at all)?

```diff
--- a/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
+++ b/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
@@ -85,6 +85,7 @@ class Dotty extends RenaissanceBenchmark {
       Thread.currentThread.getContextClassLoader
         .asInstanceOf[URLClassLoader]
         .getURLs
+        .map(p => p.toString.replace("file:/", ""))
         .mkString(File.pathSeparator),
       DOTTY_ARG_TYPE_CONVERSION,
       DOTTY_ARG_CLASS_FILE_DESTINATION,
```

vhotspur:
I have tried the suggestion from #176 discussion but that does not work as Dotty looks at the root classpath, not at the current class loader.

I have added an explanation to the source and provided a fix that converts the URLs to local file paths which should work well even on Windows.

### DIFF
--- a/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
+++ b/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
@@ -80,12 +80,34 @@ class Dotty extends RenaissanceBenchmark {
   private val DOTTY_ARG_TYPE_CONVERSION = "-language:implicitConversions"
 
   override def runIteration(c: Config): BenchmarkResult = {
+    /*
+     * Construct the classpath for the compiler. Unfortunately, Dotty is
+     * unable to use current classloader (either of this class or this
+     * thread) and thus we have to explicitly pass it. Note that
+     * -usejavacp would not work here as that reads from java.class.path
+     * property and we do not want to modify global properties here.
+     *
+     * Therefore, we leverage the fact that we know that our classloader
+     * is actually a URLClassLoader that loads the benchmark JARs
+     * from temporary directory. And we convert all the URLs to
+     * plain file paths.
+     *
+     * Note that using URLs as-is is not possible as that prepends the
+     * "file:/" protocol that is not handled well on Windows when
+     * on classpath.
+     *
+     * Note that it would be best to pass the classloader to the compiler
+     * but that seems to be impossible with current API (see discussion
+     * at https://github.com/renaissance-benchmarks/renaissance/issues/176).
+     */
+    val cp = Thread.currentThread.getContextClassLoader
+      .asInstanceOf[URLClassLoader]
+      .getURLs
+      .map(url => (new java.io.File(url.toURI)).getPath)
+      .mkString(File.pathSeparator)
     val args = Seq[String](
       DOTTY_ARG_CLASS_PATH,
-      Thread.currentThread.getContextClassLoader
-        .asInstanceOf[URLClassLoader]
-        .getURLs
-        .mkString(File.pathSeparator),
+      cp,
       DOTTY_ARG_TYPE_CONVERSION,
       DOTTY_ARG_CLASS_FILE_DESTINATION,
       outputPath.toString


benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
```diff
@@ -80,12 +80,34 @@ class Dotty extends RenaissanceBenchmark {
...
+     * thread) and thus we have to explicitly pass it. Note that
+     * -usejavacp would not work here as that reads from java.class.path
+     * property and we do not want to modify global properties here.
+     *
+     * Therefore, we leverage the fact that we know that our classloader
+     * is actually a URLClassLoader that loads the benchmark JARs
+     * from temporary directory. And we convert all the URLs to
+     * plain file paths.
+     *
+     * Note that using URLs as-is cannot be used as that prepends the
```
davleopo: "Note that using URLs as-is is not possible ..."

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
```diff
@@ -80,12 +80,34 @@ class Dotty extends RenaissanceBenchmark {
...
+     * "file:/" protocol that is not handled well on Windows when
+     * on classpath.
+     *
+     * Note that it would be best to pass the classloader to the compiler
+     * but that seems to be impossible with current API (see discussion
```
davleopo: Have you found anything about whether this may change in the future?

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
```diff
@@ -80,12 +80,34 @@ class Dotty extends RenaissanceBenchmark {
...
+     * at https://github.com/renaissance-benchmarks/renaissance/issues/176).
+     */
+    val cp = Thread.currentThread.getContextClassLoader
+      .asInstanceOf[URLClassLoader]
+      .getURLs
+      .map(url => (new java.io.File(url.toURI)).getPath)
```
davleopo: Thanks for catching this!

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
vhotspur: Thanks!

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
vhotspur: No, but I did only a quick search through the issues that are on GitHub.

Issue #143, opened 2019-05-24
Add isFinalIteration() method in Policy


Fithos:
When performing analyses, it may be useful to know when the benchmark is executing the final iteration, (e.g., to consider it a sort of "steady-state"). Currently, it seems that the `Plugin` and `Policy` classes do not offer a method such as `isFinalIteration()` to know when this is the case. It may be useful to add it.

axel22:
It seems that for some policies it might be hard to know this - for example, how would you implement this in the fixed-time policy?

lbulej:
Could you provide an example use case?

I plan for `Policy` to basically only have 2 ethods. One to determine whether to keep executing the benchmark's measured operation/repetition and one that notifies the policy that the measured operation/repetition has completed (and how long it took). This should allow a fixed-time policy to keep track of net benchmark execution time without polluting it  with execution time spent in plugin hooks.

A plugin can be notified before/after each execution of a measured operation/repetition and before/after benchmark (i.e. before the first and after the last execution of a measured operation/repetition), and thus should be able to keep track of the overall execution.

BTW, I'd suggest to discuss API requirements within issue #81 so that we can keep track of such requirements.

Fithos:
I suggested this addition because in some dynamic analyses we may want to consider all iterations apart from the last as "warm-up", avoiding to collect metrics during such iterations. Without a `isFinalMethod()`, we would need to collect metrics for all the iterations (which, depending on the analysis, may generated large traces) and then filter out metrics not collected during the last iteration. 

If this addition is in contrast to the (planned) design of the `Policy` class, then never mind. It does not prevent to execute analyses anyways :)

axel22:
@rosa-andrea Then I would propose to:

1) Have a method such as `isWarmUpIteration`, that tells you if something is a warm-up iteration or not.
2) Implement it to: in fixed-time policy to return the obvious, in fixed-iterations policy to return true for all iterations but the last.
3) If you want constant-time warmup, and one extra measurement iteration (note: fixed-time does constant time warmup and constant time measurement runs), then implement a 3rd policy that does that.

Fithos:
> @rosa-andrea Then I would propose to:
> 
> 1. Have a method such as `isWarmUpIteration`, that tells you if something is a warm-up iteration or not.
> 2. Implement it to: in fixed-time policy to return the obvious, in fixed-iterations policy to return true for all iterations but the last.
[...]

Issue #20, opened 2019-04-10
Establish a consistent way of handling the standard output across all the benchmarks


axel22:
We should ensure that all standard (output except the iteration times) goes to a separate file that can be inspected after the benchmark finishes, unless the user explicitly requests to see the benchmark output (controlled by a flag).

PR #191, opened 2019-08-02
WiP: Design/API clean-up with named configurations

lbulej:
# Introduction

This is a snapshot of the design/API clean-up mentioned in #81. Besides other things that I'm going to write about, you will notice that most of the non-benchmark classes have moved around, so to find your way around the sources, let's just say that we have 4 packages now:
+ `org.renaissance` which contains classes that are potentially interesting to benchmark writers or other developers who want to extend the harness locally through plugins or execution policies.
+ `org.renaissance.core` which contains classes that must be accessible to the Renaissance harness or the JMH wrapper. These classes live in a separate package, because they would be inaccessible to classes with the same package name loaded by a different class loader.
+ `org.renaissance.harness` which contains the Java/Scala parts of the Renaissance harness.
+ `org.renaissance.jmh` which contains the JMH wrapper

# Summary of changes

+ Support for named configurations, inspired by @vhotspur and #182 
  - Provides `@Parameter` annotation to define configurable parameters with default values and summary description.
  - Provides `@Configuration` annotation to define a named configuration which allows to override (only) configurable parameters defined using the `@Parameter` annotation.
  - Default configuration (named `default`) is automatically created for each benchmark using default parameter values.
  - A configuration without any overrides present in the `settings` attribute contains all parameters set to default values. This allows to create place-holder configurations that we want to have in all benchmarks, e.g. `@Configuration(name="test")` and `@Configuration(name="jmh")`, so that we can always run all benchmarks in these configurations.
  - The CLI allows choosing a benchmark configuration using the `-c` or `--configuration` option.
  - The JMH wrapper always uses the `jmh` configuration to query values.
  - Benchmarks can access parameter values through the `BenchmarkContext` interface, using methods such as `intParameter("name")`, `doubleParameter("name")`, or `stringParameter("name")`.
    +  I was considering a method like `Number numberParameter("name")` for numeric values so that user could call `intValue()` or `doubleValue()` on the result, but it was a bit longer to write on the client side. On the other hand, the numeric parameters would be handled by a single method. Let me know if you think it a good idea.
    + In a few places, I need to get a list of values from the parameter. For now, the benchmarks just grab the string value and parse it as needed. Let me know if you think we should include a method like `stringListParameter` or `intListParameter` or `doubleListParameter`.
  - All benchmarks have been converted to use the named configurations. I have renamed some of the variables to make it clear that it is a configurable parameter (they have the `Param` suffix). I have mostly converted only the parameters that are dynamic (thread count) or that were conditionally modified depending on the `functionalTest` flag being active.
    + The `functionalTest` CLI option and everything around it is gone, Travis runs use the `-c test` to select the testing configuration.

+ The `Policy`, `FixedIterationsPolicy`, `FixedWarmupPolicy` and `RenaissanceBenchmark` classes are gone, replaced by `ExecutionDriver`, `Plugin.ExecutionPolicy` and `ExecutionPolicies`. The benchmarks no longer have access to `Config` and have to use `BenchmarkContext` to access parameters and (in future) other harness services.
  - These have been a major source of entanglement, because the execution of a benchmark was going from the `RenaissanceSuite` object into `RenaissanceBenchmark` class, which called the `Policy`, which then called back into `RenaissanceBenchmark` which called the actual benchmark.
  - The `ExecutionDriver` now implements the whole execution sequence, only consulting `ExecutionPolicy` to determine whether to repeat the measured operation.
  - A `Plugin.ExecutionPolicy` implementation can be queried whether it is executing final iteration as per #143
  - The `ExecutionPolicies` class a collection of simple execution policies:
    + `ExecutionPolicies.FixedOpCount`, which is used by default for a fixed number of repetitions. This corresponds to the previous `FixedIterationsPolicy`, but I wanted to get rid of the word "iterations" when we agreed not to use it for the repeated execution of the measured operation and leave it for iterative algorithms instead.
    + `ExecutionPolicies.FixedOpTime`, which limits the net operation execution time (accumulates operation durations). This has no previous counterpart, but somehow it seemed a nice idea. Can be moved into a plugin and serve as an example.
    + `ExecutionPolicies.FixedTime`, which limits the execution time (counts everything from the first measured operation execution) . This almost corresponds to the old `FixedWarmupPolicy` with an important difference: it only takes a single run time value, because splitting the wall-clock execution time between "warmup" and actual "run" does not make sense. I think should avoid mentioning warmup in the harness, leaving it users to figure out.
    + Execution policy is loaded and initialized by the harness the same way as plugins (see below) and can subscribe to the same events as plugins.

+ The `BenchmarkLoader` class is gone, replaced by `BenchmarkRegistry` and `BenchmarkInfo`. 
  - The `BenchmarkRegistry` class is responsible for loading benchmark meta-data, and allows to enumerate benchmarks purely based on data loaded from a property file. Each benchmark entry has a corresponding `BenchmarkInfo` class.
  - When we want to load a benchmark, we ask the `BenchmarkRegistry` to load a benchmark (passing it a `BenchmarkInfo`) and it loads it using `ModuleLoader`. 
  - I previously had a `load()` method in the `BenchmarkInfo` class, which was kind of nice, but decided that I did not want a "descriptor class" to have to ability to load a benchmark, because the descriptor is mostly a data class.
  - The `BenchmarkRegistry` can be easily used by the RenaissanceSuite harness, JMH wrapper, and `MarkdownGenerator`.

+ The old `Plugin` is gone, replaced by a new `Plugin`, which is just a marker interface which provides a name space for a number of event listener interfaces. When a plugin wants to receive certain harness events, it should just implement the respective event listener interface and the harness will automatically register it with its event dispatcher.
  - Plugins can be loaded using a CLI parameter `--plugin`, which takes a plugin specification in the form `<class-path>!<class-name>`. This can be given multiple times to load more plugins. The last plugin loaded is the "closest" to the measured operation, from both "sides", i.e., plugins specified earlier "wrap" plugins specified later.
  - Plugins (and execution policies above) can be passed arguments using a CLI parameter `--with-arg <arg>`, which appends  <arg> to the list of arguments for the plugin that was last specified using the `--plugin` option. The `--with-arg` option can appear multiple times. If a plugin wants to receive arguments this way, it needs to implement a constructor which takes an `String[]` or `String ...` as an parameter. The harness tries to use the constructor with arguments first and falls back to using default constructor if it cannot find a constructor with arguments.
  - Plugins can publish custom measurement results. They need to implement the `MeasurementResultPublisher` interface through which they will receive a reference to a dispatcher which they can use to publish custom measurement results such as hardware counters.
  - Thanks for @vhotspur for major help and testing examples!
   
+ The various `*Result` classes are gone, moved into static factory methods in `BenchmarkResult` returning instances of anonymous classes. They turned out to be quite simple and it reduces the number of classes visible in the `org.renaissance` package. In general, we mostly want benchmark writers to either implement interfaces, or use some helpers provided by methods, not classes.

+ Black holes are gone. The assumption is that since we ask the benchmark to provide a `BenchmarkResult` on which we call `validate()` from a different context. Even though the `validate()` method does not return anything, it can actually throw an exception and I think that the easiest way for a compiler to find out whether an exception can be thrown is to execute the code. Is that a reasonable assumption?

+ The `renaissance-suite.scala` file has been split up into multiple files:
  - Generation of `README.md` and CONTRIBUTION.md has been extracted to `MarkdownGenerator.scala` which tries to limit string interpolation to hash map look-ups. The map contains values that are computed in one places to make it easier to see what values are needed and how they are obtained. The class can be run independently and is intended to be run by the build system (if someone is able to express it in SBT). For this purpose, the class accepts to command line options to provide values (renaissance title and renaissance version) that are currently loaded from the `MANIFEST.MF` at runtime, and for which the build system has the necessary values available.
  - Configuration parsing has been extracted into `ConfigParser.scala`, and the `Config` class have been converted to Scala (the Java part of the harness does not need it at all) and moved into the `org.renaissance.harness` package and is package private to the harness. This makes it easier to play with CLI changes etc.
  - Result writing has been extracted into `ResultWriter.scala`
  - The rest of the suite goes into `RenaissanceSuite.scala`, and only handles configuration processing before handing off benchmark execution to `ExecutionDriver`.

+ Method names in the `Benchmark` interface have been changed to make them a bit more streamlined and to get rid of the word "iteration" which we agreed to keep for iterative algorithms.  Instead, we should work with the concept of "measured operation" that is being repeatedly executed. @vhotspur suggested to make the names similar to what is used in testing frameworks, so I went with JUnit5 (even though it uses annotations) and renamed the methods as follows:
  - `setUpBeforeAll()` remains
  - `beforeIteration()` -> `setUpBeforeEach()`
  - `runIteration()` -> `run()`
  - `afterIteration()` -> `tearDownAfterEach()`
  - `tearDownAfterAll()`  remains

+ Assorted clean-ups / simplifications that make IDEA style warnings go away.

# Changes to consider

Another thing to consider is result validation. Right now we have the benchmark return a `BenchmarkResult` from the `run` method, which means that we are setting up the returned result in the measured operation. This is probably not big deal, because any validation activity is performed in the `validate()` method, but there is still the setup of the `BenchmarkResult` instance. An alternative is to consider returning it from `tearDownAfterEach`, which is outside the measured block. This would however require the benchmark to keep the result around until the method is called and actually require a benchmark to implement at least two methods instead of one.
- @vhotspur argued against this, supporting the notion that setting up a `BenchmarkResult` instance in `run` method is not big deal.

All thoughts are welcome.

# What does not work (yet)

In addition to custom policies and plugins mentioned above, these are the things that are not implemented and should be easier to add now:
- Harness-managed temporary directories. The `BenchmarkContext` has some placeholder methods, but in general, there should be just one (or two), where a benchmark can ask for a temporary directory which will be later deleted by the harness. This should be relatively straightforward, but needs the concept of a global scratch directory base (configurable via CLI option) into which these temporary directories will go.
- Harness-provided random generator to make sure that random values are consistent across different JDKs. However, ensuring that even large framework-based benchmarks get their random generator from us (if they use a JDK one) might require replacing the `java.util.Random` class.
- Capturing input/output from benchmarks and storing it in per-benchmark/per-repetition files. Since the execution sequence is now completely in the `ExecutionDriver`, replacing `System.out` and `System.err` around the calls to the benchmark should be fairly easy, we just need to think about where we want to store the dumps.
- Access to resources. Currently, all the benchmarks grab some class of theirs and ask for resource as a stream. Maybe the context should provide a way to grab a resource, or even unpack it into a scratch directory and provide a file name or a stream.
- Predefined event listeners (plugins) for GC or pauses before/after benchmarks/operation executions, which will be enabled simply by a CLI option. Given that there is no way to let plugins participate on the construction of the config parser, for now we would just do it the same way as with the execution policy. This is all done in the config phase, the point is that the execution driver does not need to know a thing (it only ever deals with event listeners).
 
# What fails
 
For some reason, the Travis checks fail on Mac and non-java-8 JDKs, with the following exception:
`java.lang.IncompatibleClassChangeError: Method org.renaissance.BenchmarkResult.dummy([Ljava/lang/Object;)Lorg/renaissance/BenchmarkResult; must be InterfaceMethodref constant`

I have no idea why.  

## Update 1

When I create a simple class:
```java
package org.renaissance.test;
import org.renaissance.BenchmarkResult;

public final class Test {
  public static void main (String ... args) throws BenchmarkResult.ValidationException {
    BenchmarkResult.dummy().validate();
  }
}
```
compile it using Java 8 `javac -d out BenchmarkResult.java Test.java` and run it using with Java 11 ` /usr/lib/jvm/java-11/bin/java -cp out org.renaissance.test.Test`, it works normally (and vice versa, when I compile with Java 11 and use `-source 8` and `-target 8` options).

## Update 2

When I convert `BenchmarkResult` from interface (with static methods) to abstract class (also with static methods), the problem goes away.  Is there some kind of class transformation going on behind the scenes (in Scala)? Because it seems that only the Scala benchmarks are affected. The `dummy-validation-faling` seems to work just fine (i.e., it fails as it should).

You will notice that after I committed this change to the branch, the Travis checks started to pass. I'm not too happy about the workaround, but I'm not ready to dig into Scala bugs yet.

## Update 3

The work around annoyed me, so I made `BenchmarkResult` back into an interface and created inner (name space) classes `Validators` and `Assert` to hold the static methods. This makes Scala happy again, as there are no interface methods.
 
# Conclusion

I have probably forgotten to mention something. I have tried to describe some of the things in commit messages, but there are too many commits. Documentation in classes should be up-to-date, some things made it into `README.md` and `CONTRIBUTION.md`, but some are still missing -- these will be finalized when we agree to merge this.

Please give it a try and be gentle :-)


Fithos:
Nice redesign, thank you!

renaissance-core/src/main/java/org/renaissance/core/BenchmarkInfo.java
```diff
@@ -0,0 +1,104 @@
+package org.renaissance.core;
+
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+public final class BenchmarkInfo {
+
+    final String className;
```
vhotspur: Wrong indentation. I mean, 4 spaces is the right indentation but we use 2 spaces everywhere else 😉.

benchmarks/jdk-streams/src/main/scala/org/renaissance/jdk/streams/Scrabble.scala
```diff
@@ -1,59 +1,66 @@
...
+        actualWords.size,
+        "best words count"
+      )
+
+      for ((expected, actual) <- expectedResultParam zip JavaConverters.asScalaBuffer(
+             actualWords
+           )) {
+        BenchmarkResult.assertEquals(expected, actual, "best words")
+      }
+    }
```
vhotspur: I think we should enforce the use of `return` in Scala sources. As I was switching between Java and Scala sources, my first impression was *we should not do validation inside `runIteration`*. But the missing `return` *probably* means that we return anonymous implementation of `BenchmarkResult` and the code inside `{ ... }` would be executed later. Wasn't a separate class actually more readable?

Just my two (despairing) cents.

renaissance-core/src/main/java/org/renaissance/core/BenchmarkInfo.java
lbulej: Ooops! :-)

benchmarks/jdk-streams/src/main/scala/org/renaissance/jdk/streams/Scrabble.scala
lbulej: I don't know about `return` enforcement. Leaving it out seems to be an accepted practice and I was annoyed by the various yellow markers in IDEA. If a method has uncomplicated control flow (like this one), I did not find it too confusing or annoying once I got used to Scala. However, if a method has an early return somewhere, I would probably prefer to see explicit returns on all paths.

And yes, this returns an anonymous implementation of `BenchmarkResult`. It is a bit more verbose, because this particular Scala version does not support using a lambda implementation of a SAM class (Single Anonymous Method).

As for separate class, I was removing simple classes that only serve to delay the validation computation and can be easily put into a lambda.

renaissance-core/src/main/java/org/renaissance/core/BenchmarkInfo.java
lbulej: > > * Support for named configurations, inspired by @vhotspur and #182
> 
> And where is my `dummy-params` benchmark? cry

Nothing personal, forgot about it once I converted all the benchmarks. I guess I can still add it :-)

> I would eventually like to see parameters as annotations on fields where the value would be injected. I think `intParameter` and `doubleParameter` are fine.

I think we are on the same page here. To make it easy to see what different configurations override, I would keep the `@Configuration` annotation per class (to list overrides), and use the `@Parameter` annotation per field, picking name and type from the field and allowing to set default value and description. However, not sure how we would handle list parameter types.

> 
> > * Let me know if you think we should include a method like `stringListParameter` or `intListParameter` or `doubleListParameter`.
> 
> It is already on 4 places so probably yes just to ensure that we use comma consistently everywhere.

Yeah, forgetting to trim the split values has bitten me a few times. Having a common method could have prevented that.


> > * This is just an initial sketch, tested so far only on the `ResultWriter`, which is registered to receive some of the events, allowing to remove the old `ResultObserver` interface. As we try to develop plugins, we will see whether this is sufficient.
> 
> Unless the plugin loading (if given on command-line) code is already in the pipeline, I can try to add that.

You are welcome to give it a shot.

> 
> > Not many, but I would like to change the method names int the `Benchmark` interface, so that we make them a bit more streamlined and we get rid of the word "iteration". 
> Personally, I would go with JUnit naming as that is quite familiar to Java user and use `beforeAll`, `before`, `run`, `after` and `afterAll`.

Or that. I was trying to be a bit more verbose, because the `Benchmark` is more like a wrapper for a measured operation, and when you call `benchmark.beforeAll()`, it also "runs", but not the measured operation. Only the `run` method has a verb indicating an action (in this line of reasoning, my previous `operation()` should really be `runOperation`. So I would argue to include verbs in the methods, e.g., at least setUpBeforeAll, setUp, run, tearDown, tearDownAfterAll, which is not so different from the current state, but gets rid of the word "iteration. I apologize for my OCD, I guess we are splitting hairs here :-)

> > An alternative is to consider returning it from `afterIteration/tearDownOperation`, which is outside the measured block. This would however require the benchmark to keep the result around until the method is called and actually require a benchmark to implement at least two methods instead of one.
> 
> I am against splitting this. Instantiation of a `BenchmarkResult` class should have negligible effect as the user would typically store the compute result in one field and return that class (and if that creates an unwanted overhead, it is not a benchmark we want, I guess). Any computations can be (and should be) done in the `validate()` method.

Fine with me.

> 
> > * Plugins/event listeners currently cannot publish custom results. We will need to provide some kind of callback interface that these listeners can use to publish their results (e.g., hardware counters).
> 
> For me this is the biggest issue as it prevents us to create plugin for [ubench-agent](https://github.com/D-iii-S/java-ubench-agent).

Yes, I know. But I think it should be easier to add now. If you can try to play with loading of plugins and registering them with the right listener list (in the right order), we can try to provide an interface for publishing results in one of the events.

> 
> > Please give it a try and be gentle :-)
> 
> I all for merging this in the current state.

Thanks :-)

benchmarks/jdk-streams/src/main/scala/org/renaissance/jdk/streams/Scrabble.scala
vhotspur: I think that the missing `return` is a bad (despite being accepted) practice but okay :-). We are still several characters away from PERL syntax so let's keep it like this.

vhotspur:
> > > * This is just an initial sketch, tested so far only on the `ResultWriter`, which is registered to receive some of the events, allowing to remove the old `ResultObserver` interface. As we try to develop plugins, we will see whether this is sufficient.
> > 
> > Unless the plugin loading (if given on command-line) code is already in the pipeline, I can try to add that.
> 
> You are welcome to give it a shot.

Here we go. I put the loading code into `ModuleLoader` though in the end I have not reused any code inside it.

I have used the following two plugins to mini-test it:

```java
package plugin;

import java.util.ArrayList;
import java.util.List;

import org.renaissance.Plugin;

public class ExampleOne implements Plugin,
        Plugin.OperationSetUpListener,
        Plugin.OperationTearDownListener,
        Plugin.HarnessShutdownListener {

    private static final String PREFIX = "[ONE]: ";
    
    private final List<String> results = new ArrayList<>();
    private long startNanos;
    
    public ExampleOne() {
        System.err.printf(PREFIX + "constructed\n");
    }

    @Override
    public void afterOperationSetUp(String benchmark, int opIndex, boolean isLastOp) {
        System.err.println(PREFIX + "after setup");
        startNanos = System.nanoTime();
    }
    
    @Override
    public void beforeOperationTearDown(String benchmark, int opIndex, long harnessDuration) {
        long myDuration = System.nanoTime() - startNanos;
        System.err.println(PREFIX + "before teardown");
        results.add(String.format("%s-%s: harness %d, me %d (%d)", benchmark, opIndex, harnessDuration, myDuration, myDuration - harnessDuration));
    }

    @Override
    public void beforeHarnessShutdown() {
        for (String s : results) {
            System.out.println(PREFIX + s);
        }
    }
}
```

```java
package plugin;

import java.util.ArrayList;
import java.util.List;

import org.renaissance.Plugin;

public class ExampleTwo implements Plugin,
        Plugin.OperationSetUpListener,
        Plugin.OperationTearDownListener,
        Plugin.HarnessShutdownListener {

    private static final String PREFIX = "[TWO]: ";
    
    private final List<String> results = new ArrayList<>();
    private long startNanos;
    
    public ExampleTwo() {
        System.err.printf(PREFIX + "constructed\n");
    }

    @Override
    public void afterOperationSetUp(String benchmark, int opIndex, boolean isLastOp) {
        System.err.println(PREFIX + "after setup");
        startNanos = System.nanoTime();
    }
    
    @Override
    public void beforeOperationTearDown(String benchmark, int opIndex, long harnessDuration) {
        long myDuration = System.nanoTime() - startNanos;
        System.err.println(PREFIX + "before teardown");
        results.add(String.format("%s-%s: harness %d, me %d (%d)", benchmark, opIndex, harnessDuration, myDuration, myDuration - harnessDuration));
    }

    @Override
    public void beforeHarnessShutdown() {
        for (String s : results) {
            System.out.println(PREFIX + s);
        }
    }
}
```

If you place them inside `<renaissance>/plugins/src/plugin`, then `javac -d out -cp ../target/classes/renaissance-harness/renaissance-core-0.10.0.jar src/plugin/*.java  && jar -c -f $PWD/one.jar  -C out plugin/ExampleOne.class && jar -c -f $PWD/two.jar  -C out plugin/ExampleTwo.class` creates two JARs with the plugins.

Then the following works as I would expect:

```shell
java -jar target/renaissance-gpl-0.10.0.jar \
  --plugin 'plugins/one.jar!plugin.ExampleOne' \
  --plugin 'plugins/two.jar!plugin.ExampleTwo' \
  dummy-empty -r 3
```

```
[ONE]: constructed
[TWO]: constructed
====== dummy-empty (dummy) [default], iteration 0 started ======
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
====== dummy-empty (dummy) [default], iteration 0 completed (2,211 ms) ======
====== dummy-empty (dummy) [default], iteration 1 started ======
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
====== dummy-empty (dummy) [default], iteration 1 completed (0,001 ms) ======
====== dummy-empty (dummy) [default], iteration 2 started ======
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
====== dummy-empty (dummy) [default], iteration 2 completed (0,001 ms) ======
[TWO]: dummy-empty-0: harness 2210504, me 2224400 (13896)
[TWO]: dummy-empty-1: harness 1066, me 1935 (869)
[TWO]: dummy-empty-2: harness 1038, me 1863 (825)
[ONE]: dummy-empty-0: harness 2210504, me 2784990 (574486)
[ONE]: dummy-empty-1: harness 1066, me 170658 (169592)
[ONE]: dummy-empty-2: harness 1038, me 173936 (172898)
```

I am adding the code here as a patch to avoid breaking some possible changes in this branch.

```patch
diff --git a/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java b/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
index 3f4c69f..2a01c5d 100644
--- a/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
+++ b/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
@@ -1,6 +1,7 @@
 package org.renaissance.core;
 
 import java.io.*;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.Files;
@@ -8,6 +9,9 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import org.renaissance.Plugin;
 
 public final class ModuleLoader {
   private static final URL[] URL_ARRAY_TYPE = new URL[0];
@@ -43,6 +47,44 @@ public final class ModuleLoader {
       throw new ModuleLoadingException(message, e);
     }
   }
+
+  public static Plugin loadPlugin(String spec) throws ModuleLoadingException {
+    Logger logger = Logging.getMethodLogger(ModuleLoader.class, "loadPlugin");
+    String[] parts = spec.split("!");
+    if (parts.length != 2) {
+      String message = "Plugin loading failed: expecting classpath!classname format.";
+      logger.severe(message);
+      throw new ModuleLoadingException(message);
+    }
+    String[] classpath = parts[0].split(File.pathSeparator);
+    String classname = parts[1];
+
+    /*
+     * Note that we do not use multicatch as some exceptions gives very
+     * little information about their cause (e.g. ClassNotFoundException
+     * on OpenJDK contains only the class name).
+     */
+    try {
+      URL[] classpathUrls = stringsToUrls(classpath);
+      ClassLoader parent = ModuleLoader.class.getClassLoader();
+      ClassLoader loader = new URLClassLoader(classpathUrls, parent);
+      Class<?> pluginClass = loader.loadClass(classname);
+      Object pluginObj = pluginClass.newInstance();
+      return (Plugin) pluginObj;
+    } catch (MalformedURLException e) {
+      String message = String.format("Plugin loading failed: %s.", e.getMessage());
+      logger.severe(message);
+      throw new ModuleLoadingException(message, e);
+    } catch (ClassNotFoundException e) {
+      String message = String.format("Plugin loading failed: unable to load %s (%s).", classname, e.getMessage());
+      logger.severe(message);
+      throw new ModuleLoadingException(message, e);
+    } catch (InstantiationException | IllegalAccessException e) {
+      String message = String.format("Plugin loading failed: unable to instantiate %s (%s).", classname, e.getMessage());
+      logger.severe(message);
+      throw new ModuleLoadingException(message, e);
+    }
+  }
   
   private static Map<String, String[]> getGroupJarNames(InputStream jarList) {
     Logger logger = Logging.getMethodLogger(ModuleLoader.class, "getGroupJarNames");
@@ -104,6 +146,43 @@ public final class ModuleLoader {
     return resultUrls.toArray(URL_ARRAY_TYPE);
   }
 
+  private static URL[] stringsToUrls(String[] urls) throws MalformedURLException {
+    Logger logger = Logging.getMethodLogger(ModuleLoader.class, "stringsToUrls");
+
+    /*
+     * What we do here is actually quite simple:
+     *
+     *   urls.map(p -> new URL(path)).toArray()
+     *
+     * However, URL instantiation can throw (checked!) exception that Java
+     * streams are unable to handle. So we surpress that exception and return
+     * null instead that we later filter out. This allows us to later check
+     * whether exception was thrown and artificially re-throw it.
+     *
+     * Next surprise is that it is not possible to construct URL from
+     * relative file path (which would be probably the most common scenario)
+     * so we first try full-fledged URL and then filepath URI.
+     */
+    URL[] result = Arrays.stream(urls)
+      .map(path -> {
+        try {
+          return new URL(path);
+        } catch (MalformedURLException e) {
+          try {
+            return Paths.get(path).toUri().toURL();
+          } catch (MalformedURLException e2) {
+            logger.severe(String.format("Ignoring malformed URL %s.", path));
+            return null;
+          }
+        }
+      }).filter(url -> url != null)
+      .toArray(URL[]::new);
+    if (result.length != urls.length) {
+      throw new MalformedURLException(String.format("Some URLs in %s are malformed.", String.join(File.pathSeparator, urls)));
+    }
+    return result;
+  }
+
   //
 
   public static final class ModuleLoadingException extends Exception {
diff --git a/renaissance-harness/src/main/java/org/renaissance/harness/Config.java b/renaissance-harness/src/main/java/org/renaissance/harness/Config.java
index 260b35d..c43d3d2 100644
--- a/renaissance-harness/src/main/java/org/renaissance/harness/Config.java
+++ b/renaissance-harness/src/main/java/org/renaissance/harness/Config.java
@@ -2,6 +2,8 @@ package org.renaissance.harness;
 
 import org.renaissance.Plugin;
 import org.renaissance.Plugin.*;
+import org.renaissance.core.ModuleLoader;
+import org.renaissance.core.ModuleLoader.ModuleLoadingException;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -44,16 +46,46 @@ final class Config {
   }
 
   public Config withPlugin(String pluginModule) {
-    plugins.addAll(Arrays.stream(pluginModule.split(","))
-      .map(n -> {
-        try {
-          return (Plugin) Class.forName(pluginModule).newInstance();
-        } catch (Throwable e) {
-          throw new RuntimeException(e);
-        }
-      })
-      .collect(Collectors.toList())
-    );
+    Plugin plugin = null;
+    try {
+      plugin = ModuleLoader.loadPlugin(pluginModule);
+    } catch (ModuleLoadingException e) {
+      throw new RuntimeException(e);
+    }
+
+    /*
+     * Register the plugin into respective listener lists.
+     * Note that pair events (setup and teardown) are added to opposite
+     * ends of the lists so that each plugin can wrap existing plugins.
+     *
+     * Therefore it is expected that measurement plugins would be added
+     * as the last ones on the command-line.
+     */
+    plugins.add(plugin);
+    if (plugin instanceof HarnessInitListener) {
+      harnessInitListeners.add((HarnessInitListener) plugin);
+    }
+    if (plugin instanceof HarnessShutdownListener) {
+      harnessShutdownListeners.add(0, (HarnessShutdownListener) plugin);
+    }
+    if (plugin instanceof BenchmarkSetUpListener) {
+      benchmarkSetUpListeners.add((BenchmarkSetUpListener) plugin);
+    }
+    if (plugin instanceof BenchmarkTearDownListener) {
+      benchmarkTearDownListeners.add(0, (BenchmarkTearDownListener) plugin);
+    }
+    if (plugin instanceof OperationSetUpListener) {
+      operationSetUpListeners.add((OperationSetUpListener) plugin);
+    }
+    if (plugin instanceof OperationTearDownListener) {
+      operationTearDownListeners.add(0, (OperationTearDownListener) plugin);
+    }
+    if (plugin instanceof BenchmarkResultListener) {
+      benchmarkResultListeners.add((BenchmarkResultListener) plugin);
+    }
+    if (plugin instanceof BenchmarkFailureListener) {
+      benchmarkFailureListeners.add((BenchmarkFailureListener) plugin);
+    }
 
     return this;
   }
```

> > Personally, I would go with JUnit naming as that is quite familiar to Java user and use `beforeAll`, `before`, `run`, `after` and `afterAll`.
> 
> So I would argue to include verbs in the methods, e.g., at least setUpBeforeAll, setUp, run, tearDown, tearDownAfterAll, which is not so different from the current state, but gets rid of the word "iteration. I apologize for my OCD, I guess we are splitting hairs here :-)

Sounds good :-D.

renaissance-harness/src/main/java/org/renaissance/harness/ExecutionDriver.java
```diff
@@ -0,0 +1,141 @@
...
+    bench.beforeIteration(this);
+    dispatcher.notifyAfterOperationSetUp(benchName, opIndex, isLast);
+
+    final long startNanos = System.nanoTime();
+    final BenchmarkResult result = bench.runIteration(this);
+    final long durationNanos = System.nanoTime() - startNanos;
+
+    dispatcher.notifyBeforeOperationTearDown(benchName, opIndex, durationNanos);
+    bench.afterIteration(this);
+    final long unixTsAfter = System.currentTimeMillis();
```
vhotspur: One more thing came to my mind after playing with the example plugins. Imagine user registers two plugins wrapping `runIteration`, one for hardware counters (inner) and one for `CompilationMXBean` (outer as it is more coarse-grained). Then if the hardware-counter plugin stores the data in a similar way as the current `ResultWriter`, the MX bean results would be tainted due to extra work done by the first plugin.

What I mean is that we miss callback that would allow postponing data processing in the plugins after we are outside the *possibly still measuring something* zone.

Not that we can anticipate all the possible interferences between all possible plugins but this seems to be an obvious spot (I remember the issues when adding hardware counters to DaCapo).

lbulej: Would an event after operation tear down help with this? Instead of `OperationTearDownListener`, we could have `BeforeOperationTearDownListener` and `AfterOperationTearDownListener`, with the latter providing a method with a callback parameter, which could be (optionally) used to publish measurement results.

This way a plugin could do quick things in the `beforeOperationTearDown` method (no parameter to provide results) and do other activities in the `afterOperationTearDown` method, optionally publishing measurement results.

A bit unrelated, but now that I think about it: it seems that `BenchmarkResultListener` is not a good name, because BenchmarkResult is the result returned by Benchmark), so maybe it should be renamed to MeasurementResultConsumer. In light of this, instead of `AfterOperationTearDownListener`, we could have a `MeasurementResultSupplier`, but the naming can be solved later.

So would this help? Of course, plugins could always screw things up in arbitrary fashion, but we assume responsible plugin writers (us :-)

lbulej:
> Here we go. I put the loading code into `ModuleLoader` though in the end I have not reused any code inside it.

Thanks for the patch. I have integrated it with a few changes and the test plugins still work :-)

renaissance-harness/src/main/java/org/renaissance/harness/ExecutionDriver.java
vhotspur: That sounds good and it also solves the problem how to push custom measurements from plugins to CSV/JSON.

lbulej: I have added initial support for publishing custom results. With the following code added to ExampleOne, it even produces additional results in the JSON output:
```java
@Override
public void onMeasurementResultsRequested(String benchmark, Plugin.MeasurementResultDispatcher dispatcher) {
    System.err.println(PREFIX + "measurements requested");
    dispatcher.publishMeasurementResult("dummy-one", 42);
}
```

lbulej: Now that I think about it, maybe we could use the existing MeasurementResultListener interface as the dispatcher parameter. The reason I defined a new interfaces was that I did not need the benchmark name parameter, but maybe that's not such a big deal and I would consider having one less interface a plus. What do you think?

vhotspur: The current solution is perhaps a bit cleaner but reducing amount of things to learn/remember is always better :-).

lbulej: My sentiment exactly. Nobody says that a listener cannot be a composite which just forwards the notification to its parts, which is exactly what happens here. I think one less interface is worth it, so I've made the change :-)

BTW, plugins can now take command line parameters, if they implement a `(String... args)` constructor. If they don't implement it, the default is used. So the following command  is now possible:
```
java -jar target/renaissance-gpl-0.10.0.jar \
  --json output.json \
  --plugin 'plugins/one.jar!plugin.ExampleOne' \
  --plugin 'plugins/two.jar!plugin.ExampleTwo' \
  --with-arg first-argument \
  --with-arg second-argument \
  dummy-empty -r 3
```

I know that we can pass arguments through environment variables or system properties, but this removes the need to come up with property names :-)

lbulej:
Now that @vhotspur spearheaded the loading of plugins, I have added support for loading external policies. Similar to plugins, policies can also take command line arguments, but differ from plugins in that a policy is instantiated for each benchmark.

I have updated the examples from @vhotspur to test some of the features.
The following sources should go to `plugins/src/plugin`:

`ExampleOne.java`:
```java
package plugin;

import java.util.ArrayList;
import java.util.List;

import org.renaissance.Plugin;

public class ExampleOne implements Plugin,
        Plugin.OperationSetUpListener,
        Plugin.OperationTearDownListener,
        Plugin.HarnessShutdownListener,
        Plugin.MeasurementResultPublisher {

    private static final String PREFIX = "[ONE]: ";
    
    private final List<String> results = new ArrayList<>();
    private long startNanos;
    
    public ExampleOne() {
        System.err.printf(PREFIX + "constructed\n");
    }

    @Override
    public void afterOperationSetUp(String benchmark, int opIndex, boolean isLastOp) {
        System.err.println(PREFIX + "after setup");
        startNanos = System.nanoTime();
    }
    
    @Override
    public void beforeOperationTearDown(String benchmark, int opIndex, long harnessDuration) {
        long myDuration = System.nanoTime() - startNanos;
        System.err.println(PREFIX + "before teardown");
        results.add(String.format("%s-%s: harness %d, me %d (%d)", benchmark, opIndex, harnessDuration, myDuration, myDuration - harnessDuration));
    }

    @Override
    public void beforeHarnessShutdown() {
        for (String s : results) {
            System.out.println(PREFIX + s);
        }
    }
    
    @Override
    public void onMeasurementResultsRequested(String benchmark, int opIndex, Plugin.MeasurementResultListener dispatcher) {
        System.err.println(PREFIX + "measurements requested");
        dispatcher.onMeasurementResult(benchmark, "dummy-one", 42);
    }
}
```

`ExampleTwo.java`:
```java
package plugin;

import java.util.Arrays;
import java.util.ArrayList;
import java.util.List;

import org.renaissance.Plugin;

public class ExampleTwo implements Plugin,
        Plugin.OperationSetUpListener,
        Plugin.OperationTearDownListener,
        Plugin.HarnessShutdownListener {

    private static final String PREFIX = "[TWO]: ";
    
    private final List<String> results = new ArrayList<>();
    private long startNanos;
    
    public ExampleTwo(String... args) {
        System.err.printf(PREFIX + "constructed with args: %s\n", String.join(",", Arrays.asList(args)));
    }

    @Override
    public void afterOperationSetUp(String benchmark, int opIndex, boolean isLastOp) {
        System.err.println(PREFIX + "after setup");
        startNanos = System.nanoTime();
    }
    
    @Override
    public void beforeOperationTearDown(String benchmark, int opIndex, long harnessDuration) {
        long myDuration = System.nanoTime() - startNanos;
        System.err.println(PREFIX + "before teardown");
        results.add(String.format("%s-%s: harness %d, me %d (%d)", benchmark, opIndex, harnessDuration, myDuration, myDuration - harnessDuration));
    }

    @Override
    public void beforeHarnessShutdown() {
        for (String s : results) {
            System.out.println(PREFIX + s);
        }
    }
}
```

`Policy.java`:
```java
package plugin;

import java.util.Arrays;
import java.util.ArrayList;
import java.util.List;

import org.renaissance.ExecutionPolicy;

public class Policy implements ExecutionPolicy {

    private static final String PREFIX = "[POLICY]: ";
    private final int count;
    private int executedCount;
        
    public Policy(String... args) {
        System.err.printf(PREFIX + "constructed with args: %s\n", String.join(",", Arrays.asList(args)));
        count = Integer.parseInt(args[0]);
    }

    @Override
    public boolean keepExecuting() {
        boolean result = executedCount < count;
        System.err.println(PREFIX + "keep executing? "+ result);
        return result;
    }
    
    @Override
    public void registerOperation(int index, long duration) {
        System.err.println(PREFIX + "register operation "+ index);
        executedCount = index + 1;
    }

    @Override
    public boolean isLastOperation() {
        boolean result = executedCount + 1 >= count;
        System.err.println(PREFIX + "is last operation? "+ result);
        return result;
    }
}
```

then create `plugins/out` directory and compile from the `plugins` directory:
```
javac -d out \
	-cp ../target/classes/renaissance-harness/renaissance-core-0.10.0.jar \
	src/plugin/*.java && \
jar -cf $PWD/one.jar -C out plugin/ExampleOne.class && \
jar -cf $PWD/two.jar -C out plugin/ExampleTwo.class && \
jar -cf $PWD/policy.jar -C out plugin/Policy.class
```

and finally test by running from the project root directory:
```
java -jar target/renaissance-gpl-0.10.0.jar \
  --json output.json \
  --plugin 'plugins/one.jar!plugin.ExampleOne' \
  --plugin 'plugins/two.jar!plugin.ExampleTwo' \
    --with-arg arg1 --with-arg arg2 \
  --policy 'plugins/policy.jar!plugin.Policy' \
    --with-arg 3 \
  dummy-empty dummy-param
```

You should then see output similar to this:
```
[ONE]: constructed
[TWO]: constructed with args: arg1,arg2
[POLICY]: constructed with args: 3
====== dummy-empty (dummy) [default], iteration 0 started ======
[POLICY]: is last operation? false
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-empty (dummy) [default], iteration 0 completed (0.946 ms) ======
[POLICY]: register operation 0
[POLICY]: keep executing? true
====== dummy-empty (dummy) [default], iteration 1 started ======
[POLICY]: is last operation? false
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-empty (dummy) [default], iteration 1 completed (0.001 ms) ======
[POLICY]: register operation 1
[POLICY]: keep executing? true
====== dummy-empty (dummy) [default], iteration 2 started ======
[POLICY]: is last operation? true
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-empty (dummy) [default], iteration 2 completed (0.001 ms) ======
[POLICY]: register operation 2
[POLICY]: keep executing? false
[POLICY]: constructed with args: 3
====== dummy-param (dummy) [default], iteration 0 started ======
[POLICY]: is last operation? false
[ONE]: after setup
[TWO]: after setup
thread_count: 8
meaning_of_life: 42
beyond_reason: PI
reality_check: 3.141593
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-param (dummy) [default], iteration 0 completed (1.017 ms) ======
[POLICY]: register operation 0
[POLICY]: keep executing? true
====== dummy-param (dummy) [default], iteration 1 started ======
[POLICY]: is last operation? false
[ONE]: after setup
[TWO]: after setup
thread_count: 8
meaning_of_life: 42
beyond_reason: PI
reality_check: 3.141593
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-param (dummy) [default], iteration 1 completed (0.269 ms) ======
[POLICY]: register operation 1
[POLICY]: keep executing? true
====== dummy-param (dummy) [default], iteration 2 started ======
[POLICY]: is last operation? true
[ONE]: after setup
[TWO]: after setup
thread_count: 8
meaning_of_life: 42
beyond_reason: PI
reality_check: 3.141593
[TWO]: before teardown
[ONE]: before teardown
[ONE]: measurements requested
====== dummy-param (dummy) [default], iteration 2 completed (0.393 ms) ======
[POLICY]: register operation 2
[POLICY]: keep executing? false
[TWO]: dummy-empty-0: harness 945808, me 965527 (19719)
[TWO]: dummy-empty-1: harness 1403, me 2272 (869)
[TWO]: dummy-empty-2: harness 1199, me 1845 (646)
[TWO]: dummy-param-0: harness 1016616, me 1018003 (1387)
[TWO]: dummy-param-1: harness 268917, me 269652 (735)
[TWO]: dummy-param-2: harness 393060, me 393895 (835)
[ONE]: dummy-empty-0: harness 945808, me 1208531 (262723)
[ONE]: dummy-empty-1: harness 1403, me 96720 (95317)
[ONE]: dummy-empty-2: harness 1199, me 86574 (85375)
[ONE]: dummy-param-0: harness 1016616, me 1123712 (107096)
[ONE]: dummy-param-1: harness 268917, me 374340 (105423)
[ONE]: dummy-param-2: harness 393060, me 543511 (150451)
```

vhotspur:
Here is another long snippet of source code ;-). This enables to measure hardware counters via [ubench-agent](https://github.com/D-iii-S/java-ubench-agent). The code requires some clean-up especially on error paths but the core functionality is there.

To compile this, it is necessary to have compiled the agent (`ant lib` should be sufficient) and have its `ubench-agent.jar` and `libubench-agent.so` handy. Then you need to compile the plugin with `-cp bench-agent.jar:../target/classes/renaissance-harness/renaissance-core-0.10.0.jar` and create a plugin JAR.

Then you can run the following command:

```shell
java \
    -agentpath:./libubench-agent.so \
    -jar target/renaissance-gpl-0.10.0.jar \
    --csv output.csv  --json output.json \
    --plugin 'ubench-plugin.jar:ubench-agent.jar!plugin.Main' \
    --with-arg SYS:wallclock-time,PAPI_L1_DCM \
    finagle-http -r 50
```

It collects L1 cache misses and wall-clock time. The difference between `nanos` and `SYS:wallclock-time` is basically the overhead of plugin callbacks.

```java
package plugin;

import org.renaissance.Plugin;

import java.util.List;

import cz.cuni.mff.d3s.perf.Measurement;
import cz.cuni.mff.d3s.perf.BenchmarkResults;

public class Main implements Plugin,
        Plugin.BenchmarkSetUpListener,
        Plugin.OperationSetUpListener,
        Plugin.OperationTearDownListener,
        Plugin.MeasurementResultPublisher {
    
    final int eventSet;
    
    public Main(String[] args) {
        if (args.length != 1) {
            warn("One argument should have been provided to the constructor.");
            eventSet = -1;
            return;
        }
        String[] events = args[0].split(",");
        
        eventSet = Measurement.createEventSet(1, events);
    }
    
    @Override
    public void afterBenchmarkSetUp(String benchmark) {
        Measurement.start(eventSet);
        Measurement.stop(eventSet);
    }
    
    @Override
    public void afterOperationSetUp(String benchmark, int opIndex, boolean isLastOp) {
        Measurement.start(eventSet);
    }
    
    @Override
    public void beforeOperationTearDown(String benchmark, int opIndex, long harnessDuration) {
        Measurement.stop(eventSet);
    }
    
    @Override
    public void onMeasurementResultsRequested(String benchmark, int opIndex, Plugin.MeasurementResultListener dispatcher) {
        BenchmarkResults results = Measurement.getResults(eventSet);
        String[] events = results.getEventNames();
        List<long[]> data = results.getData();
        if (data.size() != 1) {
            warn("Ignoring invalid data from this loop.");
            return;
        }
        long[] values = data.get(0);
        if (values.length != events.length) {
            warn("Ignoring invalid data from this loop.");
            return;
        }
        for (int i = 0; i < values.length; i++) {
            dispatcher.onMeasurementResult(benchmark, "ubench_agent_" + events[i], values[i]);
        }
    }
    
    private void warn(String msg, Object... args) {
        System.out.printf("[ubench plugin] WARNING: " + msg + "\n", args);
    }
}
```

lbulej:
> Here is another long snippet of source code ;-). This enables to measure hardware counters via [ubench-agent](https://github.com/D-iii-S/java-ubench-agent). The code requires some clean-up especially on error paths but the core functionality is there.

Great! We should start thinking where such a plugin would live -- whether with ubench (in which case, we should create a publishable artefact with just the API classes that can be depended on), or in a `plugins` folder of the Renaissance project, so that it is more visible to potential users, or in a separate repository (renaissance-addons) of the renaissance-benchmarks GitHub group.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
```diff
@@ -22,14 +23,22 @@ import scala.util.Random
...
+final class Als extends Benchmark with SparkUtil {
 
   // TODO: Consolidate benchmark parameters across the suite.
   //  See: https://github.com/renaissance-benchmarks/renaissance/issues/27
 
-  var numRatings = 20000
+  private var ratingCountParam: Int = _
 
-  val THREAD_COUNT = 4
+  private val THREAD_COUNT = 4
```
axel22: Should we add a parameter `thread_count` here?

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/NaiveBayes.scala
```diff
@@ -11,20 +11,19 @@ import org.apache.spark.mllib.classification.NaiveBayesModel
...
-import org.renaissance.SimpleResult
-import org.renaissance.Benchmark._
 
 @Name("naive-bayes")
 @Group("apache-spark")
 @Summary("Runs the multinomial naive Bayes algorithm from the Spark MLlib.")
 @Licenses(Array(License.APACHE2))
 @Repetitions(30)
-class NaiveBayes extends RenaissanceBenchmark with SparkUtil {
+class NaiveBayes extends Benchmark with SparkUtil {
```
axel22: Shouldn't this one also get its parameters?

benchmarks/dummy/src/main/java/org/renaissance/dummy/DummyTeardownFailing.java
```diff
@@ -1,25 +1,25 @@
...
+import org.renaissance.License;
 
 import static org.renaissance.Benchmark.*;
 
 @Name("dummy-teardown-failing")
 @Group("dummy")
 @Summary("A dummy benchmark for testing the harness (fails during teardown).")
 @Licenses(License.MIT)
-public final class DummyTeardownFailing extends RenaissanceBenchmark {
+public final class DummyTeardownFailing implements Benchmark {
```
axel22: If somebody runs all of the benchmarks together, will this fail the everything? Or just leave a note about the failure in the result files.

project/jmh.scala
```diff
@@ -8,58 +9,29 @@ object RenaissanceJmh {
...
-         defaultTearDownAfterAll();
-       }
+@State(Scope.Benchmark)
+@OutputTimeUnit(MILLISECONDS)
+@Warmup(iterations = ${info.repetitions}, time = 1, timeUnit = MILLISECONDS)
+@Measurement(iterations = ${info.repetitions / 4 + 1}, time = 1, timeUnit = MILLISECONDS)
+public class ${jmhClassName} extends JmhRenaissanceBenchmark {
+  public ${jmhClassName}() { super("${info.name}"); }
+}
+"""
```
axel22: Is `scalafmt` ok with this indentation?

project/jmh.scala
```diff
@@ -70,12 +42,18 @@ object RenaissanceJmh {
     nonGpl: Boolean,
     groupJars: Seq[(String, Seq[File], Seq[File])]
   ) = {
-    val perProjectBenchmarkClasses = for ((project, allJars, loadedJars) <- groupJars) yield {
+    val perProjectBenchmarkClasses = for {
+      (project, allJars, loadedJars) <- groupJars
+      // TODO: Filter projects in the build file if possible
+      if project.startsWith("benchmarks")
```
axel22: Maybe you want to use `"benchmarks/"` as an extra precaution.

renaissance-core/src/main/java/org/renaissance/BenchmarkContext.java
```diff
@@ -0,0 +1,67 @@
...
+  @Deprecated
+  default Path generateTempDir(String name) {
+    return DirUtils.generateTempDir(name);
+  }
+
+  @Deprecated
+  default void deleteTempDir(Path dir) {
+    DirUtils.deleteTempDir(dir);
+  }
+
```
axel22: Is another TODO to place out/err streams here? #20

axel22:
LGTM - nice redesign!

axel22:
> in a `plugins` or `addons` folder of the Renaissance project, so that it is more visible to potential users, or in a separate repository (renaissance-addons) of the renaissance-benchmarks GitHub group.

I would vote strongly for the subfolder option, i.e. in this repository.
It's easier to manage and test a single repository, rather than multiple repos.

README.md
```diff
@@ -338,6 +408,5 @@ by going through the system class loader (this can easily happen with,
 e.g. Apache Spark and Scala, due to the way that Spark internally resolves some classes).
 
 You can see the further details of the build system in the top-level `build.sbt` file,
-in the `renaissance-suite.scala` file and in `ModuleLoader`.
-
-
+and in the source code of the `RenaissanceSuite$` and
```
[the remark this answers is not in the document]
lbulej: Huh, that's what I got from `RenaissanceSuite.class.getSimpleName`. I'll just hard-code it then.

README.md
```diff
@@ -314,10 +384,10 @@ renaissance-core
 ```
 
 When the harness is started, it uses the input arguments to select the benchmark,
-and then unpacks the JARs of the corresponding benchmark group into a scratch folder.
-The harness then creates a classloader with the unpacked JARs and loads the benchmark group.
-The class loader is created directly below the default class loader. Because
-the default class loader contains only base JRE classes and common interfaces
+and then unpacks the JARs of the corresponding benchmark group into a temporary directory.
+The harness then creates a classloader which searches the unpacked JARs and loads the
```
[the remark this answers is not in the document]
lbulej: It's not discriminatory (i.e., selecting a classloader of a particular type), it just describes what the classloader is doing. But I'm fine with using "that" :-)

README.md
```diff
@@ -163,42 +167,108 @@ The following is the complete list of benchmarks, separated into groups.
...
+  with HarnessInitListener
+  with OperationSetUpListener
+  with OperationTearDownListener {
+  override def afterHarnessInit() = {
+    // Initialize the plugin after the harness finished initializing
   }
-  def beforeIteration(policy: Policy) = {
-    // Notify the tool that a benchmark iteration is about to start.
+
+  override def beforeOperation(benchmark: String, index: Int) = {
```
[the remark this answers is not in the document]
lbulej: `BenchmarkContext` is meant for benchmarks only, the plugin interface is intentionally simple at the moment. The `BenchmarkContext` interface is not really finished yet -- more methods will go in, and I'm not yet sure we want/can expose it to plugins at this point.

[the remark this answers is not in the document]
lbulej: > Does "operation" mean an "one benchmark invocation" i.e. "iteration"?
> If so, this was a bit unclear to me, maybe it should be made more obvious.

In short, yes. Long asnwer:  operation means "measured benchmark operation", i.e., the part of benchmark that is being measured. I think I mentioned it elsewhere, `benchmark.setUpBeforeAll` is technically also a benchmark invocation.

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
```diff
@@ -2,48 +2,55 @@ package org.renaissance.actors
...
+@Configurations(
+  Array(
+    new Configuration(name = "test", settings = Array("loop_count = 2")),
+    new Configuration(name = "jmh", settings = Array("loop_count = 1"))
+  )
+)
+final class AkkaUct extends Benchmark {
 
   // TODO: Consolidate benchmark parameters across the suite.
   //  See: https://github.com/renaissance-benchmarks/renaissance/issues/27
```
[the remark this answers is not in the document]
lbulej: I'm not sure. It was you who opened it :-)

This PR provides support for named configurations and (at the moment) its supports one configuration tag (`$cpu.count`) which expands to number of available CPUs.

But does it count as having consolidated the benchmark parameters? If yes, then I can remove the comments, otherwise it should be removed on a case-by-case basis.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/NaiveBayes.scala
lbulej: I think I only added parameters to benchmarks which explicitly checked for "test" run and thus created a need for a different configuration. For other benchmarks, there are certainly candidates for configurable parameters, but some benchmarks have many of them and I could not tell which ones are worth configuring and which ones not.

benchmarks/dummy/src/main/java/org/renaissance/dummy/DummyTeardownFailing.java
lbulej: It would only fail the one benchmark but dummy benchmarks are excluded from the "all" group.

project/jmh.scala
lbulej: As far as I can tell, it is :-) I don't actually see what's wrong with it when it is in a string.

project/jmh.scala
lbulej: Will do.

renaissance-core/src/main/java/org/renaissance/BenchmarkContext.java
lbulej: That's something to decide. For "our" benchmark code, it will work fine. But many of the benchmarks actually embed gobs of other code which will not use our out/err from the `BenchmarkContext`.

One options is for the harness to use `System.setOut` and `System.setErr` around invocations of `Benchmark` methods -- then we would not need to provide out/err in `BenchmarkContext`.

An alternative is to provide them and have "our" benchmark wrappers set it if they use gobs of foreign code. I'm open to suggestions here.

README.md
```diff
@@ -163,42 +167,108 @@ The following is the complete list of benchmarks, separated into groups.
...
+The suite generally executes the benchmark's measured operation multiple times
+and uses an execution policy to determine how many times to repeat the execution.
+By default, the suite supports executing the measured operation for a fixed number
+of times and for a fixed amount of time. These policies are implicitly selected by
+setting the number of iterations or the execution time using command line options.
 
-- `fixed-iterations` -- Runs the benchmark for a fixed number of iterations (use `-r`).
+To provide additional control over execution of the measured operation, the suite
+allows using an external execution policy. An external policy is a user-defined class
+which implements the ExecutionPolicy interface and provide at least a default
```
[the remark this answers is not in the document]
lbulej: I have updated the documentation after the latest changes which treat execution policy as a plugin, so it is not necessary anymore.

lbulej:
OK, one more design change and that I'll drop the bone for a while. 

I was annoyed by the need to handle execution policies in a special way. In contrast to "normal" plugins, we needed to create an instance of execution policy for each benchmark. This made it a little easier for execution policy to maintain state, but it made the harness more complicated. When I wanted to add support for registering to other plugin events (so that a policy could peek, e.g., at custom results), it was quite painful, so after a while I scrapped the effort and made execution policy into a normal plugin.

This moves the responsibility for maintaining per-benchmark state to the policy itself, but since the policy can register for all other events, it's not actually that difficult and none of the code is on performance critical part anyway. The harness can treat the policy like normal plugin, especially when it comes to registering it for events.

The policy still needs to be specified using the `--policy` flag to make it clear from the command line that a custom policy is being used (the alternative was to just use `--plugin` for everything and have the harness look for plugins implementing the `ExecutionPolicy` interface and pick the first/last found).

Still, the position of the `--policy` option with respect to other `--plugin` options is important and determines "how far" from the measured operation the policy will be as a listener. Because it usually does not need precise timing information, it should generally go first. If you feel that this is too sharp a knife to give to users, I can special case it so that policy will be always the first/last listener in pair events, and the last in non-pair events.

Here is an updated example of a policy which has to handle it's own per-benchmark state (not that it actually has to do much):
```java
package plugin;

import java.util.Arrays;
import java.util.ArrayList;
import java.util.List;

import org.renaissance.Plugin;

public class Policy implements
    Plugin.ExecutionPolicy,
    Plugin.BenchmarkSetUpListener, 
    Plugin.OperationTearDownListener {

    private static final String PREFIX = "[POLICY]: ";
    private final int countLimit;
    private int count;
        
    public Policy(String... args) {
        System.err.printf(PREFIX + "constructed with args: %s\n", String.join(",", Arrays.asList(args)));
        countLimit = Integer.parseInt(args[0]);
    }

    @Override
    public void afterBenchmarkSetUp(String benchName) {
        System.err.println(PREFIX + "reset on new benchmark: "+ benchName);
        count = 0;
    }

    @Override
    public void beforeOperationTearDown(String benchName, int opIndex, long durationNanos) {
        System.err.println(PREFIX + "finished operation "+ opIndex);
        count = opIndex + 1;
    }

    @Override
    public boolean canExecute(String benchName, int opIndex) {
        boolean result = count < countLimit;
        System.err.println(PREFIX + "keep executing? "+ result);
        return result;
    }

    @Override
    public boolean isLast(String benchName, int opIndex) {
        boolean result = count + 1 >= countLimit;
        System.err.println(PREFIX + "is last operation? "+ result);
        return result;
    }
}
```
The built-in policies can be found in `ExecutionPolicies` and as even simpler.

Anyway, when compiled and run with other plugins as follows:
```
java -jar target/renaissance-gpl-0.10.0.jar \
  --csv output.csv --json output.json \
  --policy 'plugins/policy.jar!plugin.Policy' \
    --with-arg 1 \
  --plugin 'plugins/one.jar!plugin.ExampleOne' \
  --plugin 'plugins/two.jar!plugin.ExampleTwo' \
    --with-arg arg1 --with-arg arg2 \
  dummy-empty dummy-param
```

the harness produces the following output:
```
[POLICY]: constructed with args: 1
[ONE]: constructed
[TWO]: constructed with args: arg1,arg2
[POLICY]: reset on new benchmark: dummy-empty
[POLICY]: keep executing? true
====== dummy-empty (dummy) [default], iteration 0 started ======
[POLICY]: is last operation? true
[ONE]: after setup
[TWO]: after setup
[TWO]: before teardown
[ONE]: before teardown
[POLICY]: finished operation 0
[ONE]: measurements requested
====== dummy-empty (dummy) [default], iteration 0 completed (1.593 ms) ======
[POLICY]: keep executing? false
[POLICY]: reset on new benchmark: dummy-param
[POLICY]: keep executing? true
====== dummy-param (dummy) [default], iteration 0 started ======
[POLICY]: is last operation? true
[ONE]: after setup
[TWO]: after setup
thread_count: 8
meaning_of_life: 42
beyond_reason: PI
reality_check: 3.141593
[TWO]: before teardown
[ONE]: before teardown
[POLICY]: finished operation 0
[ONE]: measurements requested
====== dummy-param (dummy) [default], iteration 0 completed (2.218 ms) ======
[POLICY]: keep executing? false
[TWO]: dummy-empty-0: harness 1593272, me 1614957 (21685)
[TWO]: dummy-param-0: harness 2218270, me 2219911 (1641)
[ONE]: dummy-empty-0: harness 1593272, me 1799302 (206030)
[ONE]: dummy-param-0: harness 2218270, me 2417496 (199226)
```

README.md
```diff
@@ -163,42 +167,108 @@ The following is the complete list of benchmarks, separated into groups.
...
+The events supported by the harness are represented by the following interfaces:
+- `HarnessInitListener`
+- `HarnessShutdownListener`
+- `BenchmarkSetUpListener`
+- `BenchmarkTearDownListener`
+- `MeasurementResultListener`
+- `MeasurementResultPublisher`
+- `BenchmarkFailureListener`
+- `OperationSetUpListener`
+- `OperationTearDownListener`
```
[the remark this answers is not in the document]
lbulej: `BenchmarkContext` was intended for benchmarks and unless there is an actual need, I would not give it to plugins.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
lbulej: I don't have a problem with that. As I mentioned elsewhere, I only added configurable parameters for things that were being changed in response to "functionalTest" check.

lbulej:
> LGTM - nice redesign!

Thanks! So far no one is screaming in horror, so I guess it can go in :-)

benchmarks/actors/src/main/scala/org/renaissance/actors/AkkaUct.scala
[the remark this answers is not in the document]
farquet: I would say that consolidating the parameters also means defining reasonable configurations for each benchmark and possibly also changing the default values. I would keep track of this, either in #27 or in another issue.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
```diff
@@ -74,17 +88,22 @@ class PageRank extends RenaissanceBenchmark with SparkUtil {
       .distinct()
       .groupByKey()
       .cache()
-    ranks = links.mapValues(v => 1.0)
+
+//    ranks = links.mapValues(v => 1.0)
```
davleopo: Left over?

renaissance-core/src/main/java/org/renaissance/Benchmark.java
```diff
@@ -27,64 +39,185 @@
...
+  @Target(ElementType.TYPE)
+  @Repeatable(Parameters.class)
+  @Retention(RetentionPolicy.RUNTIME)
+  @interface Parameter {
+    String name();
+    String defaultValue() default "";
+    String summary() default "";
+  }
+
+  //
```
davleopo: left over?

renaissance-core/src/main/java/org/renaissance/BenchmarkResult.java
```diff
@@ -1,5 +1,208 @@
...
+  /**
+   * Validates this benchmark result. Benchmarks are not expected to fail,
+   * but if the result is invalid, this method must throw an exception with
+   * a message providing details on why the validation failed.
+   *
+   * @throws ValidationException if the result is invalid.
+   */
+  void validate() throws ValidationException;
+
+  //
```
davleopo: left over?

renaissance-core/src/main/java/org/renaissance/BenchmarkResult.java
```diff
@@ -1,5 +1,208 @@
...
+
+      return () -> {
+        for (final BenchmarkResult result : results) {
+          result.validate();
+        }
+      };
+    }
+  }
+
+  //
```
davleopo: left over?

renaissance-core/src/main/java/org/renaissance/BenchmarkResult.java
```diff
@@ -1,5 +1,208 @@
...
+    ) throws ValidationException {
+      if (!expected.equals(actual)) {
+        throw new ValidationException(
+          "%s: expected %s but got %s", subject, expected, actual
+        );
+      }
+    }
+  }
+
+  //
```
davleopo: left over?

renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
```diff
@@ -0,0 +1,225 @@
...
+      }
+    };
+
+    return Arrays.stream(urls)
+      .map(makeUrl)
+      .filter(Objects::nonNull)
+      .toArray(URL[]::new);
+  }
+
+  //
```
davleopo: left over?

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
lbulej: This one yes :-)

renaissance-core/src/main/java/org/renaissance/Benchmark.java
lbulej: This one is intentional (and all the others except the one with commented out line) -- I use empty EOL comments to separate blocks of related methods. I can stop doing that if people object.

Issue #194, opened 2019-08-22
Specify plugin class inside plugin JAR


vhotspur:
When creating and using a plugin, it is added to the command-line via

```shell
--plugin "path-to-plugin.jar:other-deps.jar!org.package.ClassWithPlugin"
```

which can get quite long to type (and shell offers no completion for the classname). Furthermore, the exclamation mark has a special meaning inside shell when not escaped.

We should either abuse `Main-Class` to store the main class of the plugin or add some other field to `MANIFEST.MF` inside the plugin JAR that could be picked automatically by the extension loader.

lbulej:
I think that we could agree on looking for `Renaissance-Plugin` attribute in the JAR manifests when the plugin specifier does not contain any `!`. I would avoid abusing `Main-Class`.

Even though we (me and @vhotspur ) have previously discussed the option of having some plugins packaged with the harness (and activating them just by using the class name, also without `!`), I think that it is not really needed that much -- if there is (optional) functionality that we think is worth including with the harness, we should just activate it with a separate CLI option.

PR #195, opened 2019-08-22
ubench-agent plugin [WiP]

vhotspur:
This PR adds a new plugin to enable measurement of hardware counters via [ubench-agent](https://github.com/D-iii-S/java-ubench-agent). The code was originally mentioned in #191.

The plugin is compiled via SBT but user has to first compile the C agent.

After that, following spell runs Renaissance and collects L1 cache misses and wall-clock time.

```shell
java \
    -agentpath:plugins/ubench-agent/libubench-agent.so \
    -jar target/renaissance-gpl-0.10.0.jar \
    --csv output.csv  --json output.json \
    --plugin 'plugins/ubench-agent/target/plugin-ubenchagent-assembly-0.0.1.jar!org.renaissance.plugins.ubenchagent.Main' \
    --with-arg SYS:wallclock-time,PAPI_L1_DCM \
    finagle-http -r 50
```

There are few things to decide/discuss.

* Do we want the plugins in the same repository? Personally, I would prefer separate repositories as it seems cleaner (and more flexible wrt. to updates) but I have no strong opinion on this.
* How do we build dependencies for the plugins? For ubench-agent, we need to build the native agent (the JAR can be probably uploaded to Maven central or similar, but I doubt it would work well for the native libraries). Currently, there is a simple shell script that clones the repository. Not sure if Git submodule would help here.
* I have already opened #194 as the command-line is really long.

### DIFF
--- a/plugins/ubench-agent/build-ubench-agent.sh
+++ b/plugins/ubench-agent/build-ubench-agent.sh
@@ -6,13 +6,16 @@ msg() {
     echo "[build-ubench-agent]:" "$@" >&2
 }
 
-UBENCH_AGENT_DIR="ubench-agent-repo"
+UBENCH_AGENT_COMMIT="5e8473dbc4d38948dc2d680b5fdf0775824ad40f"
+UBENCH_AGENT_TARBALL="java-ubench-agent-$UBENCH_AGENT_COMMIT.tar.gz"
+UBENCH_AGENT_TARBALL_URL="https://github.com/D-iii-S/java-ubench-agent/archive/$UBENCH_AGENT_COMMIT.tar.gz"
+UBENCH_AGENT_DIR="java-ubench-agent-$UBENCH_AGENT_COMMIT"
 
 msg "Fetching sources..."
+wget --continue -O "$UBENCH_AGENT_TARBALL" "$UBENCH_AGENT_TARBALL_URL"
 if ! [ -d "$UBENCH_AGENT_DIR" ]; then
-    git clone https://github.com/D-iii-S/java-ubench-agent.git "$UBENCH_AGENT_DIR"
-else
-    ( cd "$UBENCH_AGENT_DIR"; git pull )
+    msg "Unpacking the tarball..."
+    tar xzf "$UBENCH_AGENT_TARBALL"
 fi
 
 msg "Building agent..."
--- a/plugins/ubench-agent/build-ubench-agent.sh
+++ b/plugins/ubench-agent/build-ubench-agent.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+set -ueo pipefail
+
+msg() {
+    echo "[build-ubench-agent]:" "$@" >&2
+}
+
+UBENCH_AGENT_DIR="ubench-agent-repo"
+
+msg "Fetching sources..."
+if ! [ -d "$UBENCH_AGENT_DIR" ]; then
+    git clone https://github.com/D-iii-S/java-ubench-agent.git "$UBENCH_AGENT_DIR"
+else
+    ( cd "$UBENCH_AGENT_DIR"; git pull )
+fi
+
+msg "Building agent..."
+pushd "$UBENCH_AGENT_DIR"
+ant lib
+msg "Build succeeded."
+popd
+
+AGENT_NATIVE_FILE=`find $UBENCH_AGENT_DIR/out/lib -name '*agent*' | grep -v '\.jar$'`
+
+msg "Copying agent files..."
+cp -f "$AGENT_NATIVE_FILE" .
+cp -f "$UBENCH_AGENT_DIR/out/lib/ubench-agent.jar" lib
+
+AGENT_NATIVE_FILE=$( realpath $( basename $AGENT_NATIVE_FILE ) )
+
+msg "Add the following to Java command line when starting the suite:"
+msg "  -agentpath:$AGENT_NATIVE_FILE"
--- a/plugins/ubench-agent/src/main/java/org/renaissance/plugins/ubenchagent/Main.java
+++ b/plugins/ubench-agent/src/main/java/org/renaissance/plugins/ubenchagent/Main.java
@@ -1,0 +1,68 @@
+package org.renaissance.plugins.ubenchagent;
+
+import org.renaissance.Plugin;
+
+import java.util.List;
+
+import cz.cuni.mff.d3s.perf.Measurement;
+import cz.cuni.mff.d3s.perf.BenchmarkResults;
+
+public class Main implements Plugin,
+    Plugin.BenchmarkSetUpListener,
+    Plugin.OperationSetUpListener,
+    Plugin.OperationTearDownListener,
+    Plugin.MeasurementResultPublisher {
+
+  final int eventSet;
+
+  public Main(String[] args) {
+    if (args.length != 1) {
+      warn("One argument should have been provided to the constructor.");
+      eventSet = -1;
+      return;
+    }
+    String[] events = args[0].split(",");
+
+    eventSet = Measurement.createEventSet(1, events);
+  }
+
+  @Override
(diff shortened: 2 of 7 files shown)


plugins/ubench-agent/src/main/java/org/renaissance/plugins/ubenchagent/Main.java
```diff
@@ -0,0 +1,68 @@
...
+
+  final int eventSet;
+
+  public Main(String[] args) {
+    if (args.length != 1) {
+      warn("One argument should have been provided to the constructor.");
+      eventSet = -1;
+      return;
+    }
+    String[] events = args[0].split(",");
```
lbulej: The code here could be more benevolent, accepting events from multiple arguments :-)
It would allow more flexible formatting of the command line using the plugins.

vhotspur: Right, fixed in c6acae8d3cce117ddc396078d0e7ea104e632a4a.

plugins/ubench-agent/build-ubench-agent.sh
```diff
@@ -0,0 +1,33 @@
...
+
+msg() {
+    echo "[build-ubench-agent]:" "$@" >&2
+}
+
+UBENCH_AGENT_DIR="ubench-agent-repo"
+
+msg "Fetching sources..."
+if ! [ -d "$UBENCH_AGENT_DIR" ]; then
+    git clone https://github.com/D-iii-S/java-ubench-agent.git "$UBENCH_AGENT_DIR"
```
farquet: We should stick to a specific commit/tag/release when cloning

lbulej:
Regarding repositories and submodules... I would lean towards another option.

Let plugins live in separate repositories (plugins developed by us would live in the `renaissance-benchmarks` group here at GitHub) and link them as GIT modules (referencing specific commit ids) in the `plugins` subdirectory.  

This way the plugins could live separate lives, yet those that we endorse/use/know would be linked to the main repository as a module. Before a versioned (source) release of Renaissance, we would update the GIT modules to point at a specific commit in a particular plugin's repository. This way we could also easily link to (endorsed) plugins from elsewhere (I don't expect there to be that many).

This might add a little "transactional cost", but I assume plugins will be generally built and packaged separately from the main bundle (they are optional). Given that the submodules will reference a specific commit id, we can still bundle some of them into the Renaissance bundle if we want, but their development timeline will be decoupled from Renaissance.

axel22:
We use the multiple repo approach in various projects because we have to (due to various IP issues and licensing issues).
However, it's not pretty, and it results in a lot of overhead - everybody on our team would be happier if we didn't have to do that. The PR process is made more complicated. We had to implement extra infrastructure to merge multiple PRs across different repositories at the same time. The PR process takes a much longer amount of time as a result of this. The effort required to do all this was quite big. And despite the effort to make this smooth, we often get transient failures.

If there are plugins developed exclusively for Renaissance, then I would recommend that they live in the same repo. I recommend that only if there are external plugins that have to be maintained separately, then they are depended on in a classic way - as an external binary package that is hosted somewhere else.

lbulej:
I'm not sure if the context in which you had (negative) experience with multi-repo setup is applicable to Renaissance. 

What I mean by that is that Renaissance hardly reaches the complexity of your other projects, the interface between the plugins and Renaissance is simple (after all, it's just about hooking a plugin into the execution sequence), and most importantly, Renaissance does not depend on external plugins for anything (built-in functionality that uses the same extension mechanism as plugins is a different matter).

In my view, whatever external plugins do is not our business, and if I (personally) develop a plugin, I'll be doing it wearing a different hat (that of a plugin developer depending on Renaissance API) and should not need to involve Renaissance maintainers when I commit changes to the plugin repository.

If Renaissance maintainers decide to endorse a particular plugin (typically because they would be using it) by linking to it as a GIT module in the `plugins` directory, it means that they assume certain responsibility for maintaining a link to a working version (commit id) of the plugin. They should follow updates in the plugin project and occasionally bump the GIT module commit id. If a particular plugin developer also happens to be a Renaissance maintainer, it's a bonus, but I think the development processes of the Renaissance suite and that of plugins should be separated.

Anyway, this is my purist view of the matter, so let's consider the example of the `ubench-agent` *plugin*. The agent has its own repository and can serve multiple projects, so what we are really discussing is whether the glue code between Renaissance and the agent is going to be part of the Renaissance code base or not. Because the glue code is not the agent itself, there is a script that fetches a specific commit id (as requested even on this PR) from the agent project, which is basically multi-repo setup anyway. Except the dependency is hard-coded in a script and not in a GIT module description.

In the end, it is not much different from what I'm proposing -- I'm just suggesting a different cut-off point to keep Renaissance and plugins loosely coupled.

farquet:
I agree with @lbulej since I can also see the interest of keeping it separate such that other benchmark suites could more easily use it. In the interest of the community, we could keep it separate.
Moreover, I don't think it's a big deal wrt maintenance and development to have it in a separate repo. I don't see how different this is from any other dependency, like the dozens of maven dependencies we have across all benchmarks. We stick to a specific version and if someone wants to update it, it is that person's responsability.
I would be against this if the agent would need frequent updates though since this is an extra step on the renaissance side, but that seems quite unlikely.

axel22:
@farquet If the plugin is specifically developed for Renaissance, then the other benchmark suites can still use them by checking out Renaissance, and implementing an appropriate adapter (for example, DaCapo would have to implement its Callback class).
In the case of plugins that are merely wrappers around standalone tools maintained elsewhere, other benchmark suites would anyway not benefit from Renaissance-specific plugins (which would be wrappers around the standalone tool).
So, that's not a very strong argument.

Regarding the maintenance - it is a big deal, because it takes you more time to merge a set of cross-repo changes. I'm talking here about potential plugins that we develop for Renaissance specifically. You have to first update the repo with the plugin, then go back and update the Renaissance repo. If you have tests or benchmark jobs for the plugin in the Renaissance repo, then you risk breaking the Renaissance repo in the interim.

It's not impossible, but it's extra work to implement the infra for cross repo work, and it's more time to work with such an infra.

The difference between e.g. maven dependencies and what we have here is that we will be developing both Renaissance and the plugin/agent at the same time (in particular when it comes to plugins that we specifically implement for Renaissance). The plugin will depend on the API, and Renaissance might have jobs (tests or benchmarks) that depend on the plugin. So, it's a tighter, two-way dependency, whereas with Maven you just occasionally bump the version when someone (independently of you) releases a newer library version that you decide to depend on.

Regarding your comment that it's unlikely that the agent would need frequent updates - it always seems unlikely until it turns out it's not.

@lbulej Regarding the modules dependency - I'd say that if a plugin is a standalone tool, and Renaissance repo 
implements the glue code, i.e. our plugin interface, then it's fine to use git modules (or just check out the target repo as part of the build process). But I advise strongly against having the glue code outside of this repo.

vhotspur:
Okay, my $0.02:

Let's publish Renaissance API JAR (i.e. the JAR with classes from `renaissance-core`) to Maven central. Plugins would then depend upon this published JAR and can live outside main repository.

This ensures that all plugins would be developed the same way (i.e. without touching Renaissance sources). You would not need to recompile Renaissance to develop a new plugin.

Plugins endorsed by Renaissance team would simply live in `renaissance-benchmarks` organization.

The plugins could use latest release of Renaissance for testing without need to recompile everything.

This also gives the developers freedom how to manage the plugin - whether the plugin would be a glue code (as for ubench-agent) or it extends existing codebase (i.e. as if we added the plugin to ubench-agent repository) etc. as there would be no difference regarding compilation/testing.

Regarding testing of existing plugins together with Renaissance development: currently it takes about 1/2 hour to compile and run the benchmarks on Travis (for one configuration). Adding more options to the testing matrix would probably cause running out of allocated time. Transferring this responsibility to plugins authors could help solve that.

Issue #193, opened 2019-08-13
Force GC between iterations


farquet:
The harness should force GC between iterations by default and provide an option to disable that feature.

Issue #107, opened 2019-05-03
Modify timing information in CSV/JSON output (PR #98)


ceresek:
For some processing it is useful to have information on absolute time of benchmark repetitions (start + end), for example to see how much time is spent between repetitions (GC or validation), or to correlate with events reported by the VM (GC log, JIT log), which have timestamps measured since VM start. I would suggest the following modifications:

- At the start of the harness, acquire delta between System.nanoTime and VM start as reported by java.lang.Runtime MX bean (= nanoStartTime).
- For each benchmark repetition, report time when the repetition started relative to VM start (nanoTime-nanoStartTime) and repetition duration (as reported now).
- In JSON, and in comments in CSV header, include information on VM start as ISO UTC timestamp (to permit correlation with external logs).

ceresek:
@lbulej I was thinking a bit about data size - the VM start thing is a constant that would be repeated after each iteration, so a bit of a waste (easily compressed of course).

vhotspur:
@ceresek Seems to me that this was fixed via #216 and #218. Can we close this?

PR #196, opened 2019-08-26
WiP: Force GC before measured operation by default

lbulej:
This provides initial support for forced GC mentioned in #193.

The CLI provides a `--force-gc <when>` option which allows specifying `before`, `after` or `around` in place of `<where>`. The option can be used multiple times, so using `--force-gc before` and `--force-gc after` at the same time has cumulative effect. 

While this is a simple feature, there are a couple of things that should be decided explicitly and which may trigger some other changes -- feedback is therefore welcome.

It is internally implemented as a built-in, lowest-priority plugin (executes first/last in paired notifications) hooked either into the `afterOperationSetUp` (i.e. before operation) or the `beforeOperationTearDown` (i.e. after operation) notification slots. It is also inside the unix-timestamped phase covering operation set up, plugin invocation, the actual operation, plugin invocation (again), and operation tear down. Currently it also currently prints a message *inside* the measured operation bracket (delimited by `=====`). 

Anything described above can be challenged and changed. I would bring your attention to the following:

+ Are we fine the before/after semantics and the position in the execution sequence? 
  - The *before operation* GC executes after operation setup and before operation, so that it can potentially clean garbage produced during operation setup, along with garbage from the previous operation and operation tear down.
  - The *after operation* GC raises a few more questions. If enabled, it executes after operation and before operation tear down, so it may trigger clean up of the "operation garbage" but not "operation tear down garbage".  If used alone, it will also not clean the "operation set up" garbage. 

+ Do we actually need the "force GC after operation" mode?
  - If yes, should it be moved after benchmark tear down? This may require creating an "after operation tear down" notification slot and move it there, or special case it in some other way.

+ Do we want the forced GC to print any message? If yes, when and with what content?

+ Should the forced GC be part of the unix-stamped phase of the execution sequence?

To make the discussion a bit easier, here's the execution sequence for each benchmark after it is instantiated (for reference):

1. call `setUpBeforeAll()` on the benchmark
2. notify plugins in the `afterBenchmarkSetUp` slot
3. repeat executing the measured operation while execution policy says so

    3.1 print `=====` bracket with operation started information

    3.2 take `currentTimeMillis()` (unix) timestamp of the start of the operation-related activities

    3.3 call `setUpBeforeEach()` on the benchmark

    3.4 notify plugins in the `afterOperationSetUp` slot

    - This is where forced GC and built-in policy plugins get executed, *before* all other plugins. 

    3.5 execute the measured operation

      - take `nanoTime()` timestamp of the start of the actual operation  
      - call `run()` on the benchmark
      - take `nanoTime()` timestamp of the end of the actual operation

    3.6 notify plugins in the `beforeOperationTearDown` slot

    - This is where forced GC and built-in policy plugins get executed, *after* all other plugins.

    3.7 call `tearDownAfterEach()` on the benchmark

    3.8 take `currentTimeMillis()` (unix) timestamp of the end of the operation-related activities

    3.9 validate benchmark result

    3.10 publish measured values
    - operation duration (difference between `nanoTime()` timestamps)
    - both phase start and phase end `currentTimeMillis()` (unix) timestamps

    3.11 request measurements from plugins in the `measurementResultsRequested` slot

    3.12 print `=====` bracket with operation completed information

4. notify plugins in the `beforeBenchmarkTearDown` slot
5. call `tearDownAfterAll()` on the benchmark

### DIFF
--- a/CONTRIBUTION.md
+++ b/CONTRIBUTION.md
@@ -31,10 +31,10 @@ import org.renaissance.Benchmark._
 
 @Summary("Runs some performance-critical Java code.")
 final class MyJavaBenchmark extends Benchmark {
-  override def runIteration(config: Config): BenchmarkResult = {
+  override def run(context: BenchmarkContext): BenchmarkResult = {
     // This is the benchmark body, which in this case calls some Java code.
     JavaCode.runSomeJavaCode()
-    // Return object for later validation of the iteration.
+    // Return object for later validation of the operation result.
     return new MyJavaBenchmarkResult()
   }
 }
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Usage: renaissance [options] [benchmark-specification]
   --json <file-path>       Output results to JSON file.
   -c, --configuration <name>
                            Run benchmarks with given named configuration.
-  --force-gc <when>        Force garbage collection 'before', 'after', or 'around' measured operation.
+  --no-forced-gc           Do not force garbage collection before each measured operation.
   --list                   Print list of benchmarks with their description.
   --raw-list               Print list of benchmarks (each benchmark name on separate line).
   --group-list             Print list of benchmark groups (each group name on separate line).
@@ -207,9 +207,10 @@ to provide. This is demonstrated in the following example:
 
 ```scala
 class SimplePlugin extends Plugin
-  with HarnessInitListener
-  with OperationSetUpListener
-  with OperationTearDownListener {
+  with AfterHarnessInitListener
+  with AfterOperationSetUpListener
+  with BeforeOperationTearDownListener {
+
   override def afterHarnessInit() = {
     // Initialize the plugin after the harness finished initializing
   }
@@ -226,13 +227,14 @@ class SimplePlugin extends Plugin
 
 The following interfaces provide common (paired) event types which allow a plugin to hook
 into a specific point in the benchmark execution sequence. They are analogous to common
-annotations known from testing frameworks such as JUnit.
-- `HarnessInitListener`
-- `HarnessShutdownListener`
-- `BenchmarkSetUpListener`
-- `BenchmarkTearDownListener`
-- `OperationSetUpListener`
-- `OperationTearDownListener`
+annotations known from testing frameworks such as JUnit. Harness-level events occur only
+once per the whole execution, benchmark-level events occur once for each benchmark
+executed, and operation-level events occur once for each execution of the measured
+operation.
+- `AfterHarnessInitListener`, `BeforeHarnessShutdownListener`
+- `BeforeBenchmarkSetUpListener`, `AfterBenchmarkTearDownListener`
+- `AfterBenchmarkSetUpListener`, `BeforeBenchmarkTearDownListener`
+- `AfterOperationSetUpListener`, `BeforeOperationTearDownListener`
 
 The following interfaces provide special non-paired event types:
 - `MeasurementResultListener`, intended for plugins that want to receive
@@ -244,7 +246,7 @@ has either failed in some way (the benchmark triggered an exception), or that th
 operation produced a result which failed validation. This means that no measurements results
 will be received.
 
-And finally the following interface are used by the harness to request
+And finally the following interfaces are used by the harness to request
 services from plugins:
 - `MeasurementResultPublisher`, intended for plugins that want to collect
 values of additional metrics around the execution of the benchmark operation. The harness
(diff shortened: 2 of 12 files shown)


davleopo:
Hi, thx for the feature!

Code changes LGTM, however policy wise I am not sure we want the separation of before and after.
What is the rationale to both ?
It makes it a bit more complex I think then it needs to be, but maybe I am missing a use-case in my head :-)
Regarding printing I would vote for keeping the printing that way everybody sees that gc is invoked, makes it more explicit.

ceresek:
IIRC the reason for having both 'before' and 'after' was tied to our earlier measurements where we computed allocation rate by adding up 'total memory freed by GC' for all repetitions. If no GC was done after the last repetition, the total was off by that one repetition (and if GC were done only 'after', the total would be off by the initial allocations).

ceresek:
As far as the various timestamps go, see #107 ? I don't know the exact reason for the way the timestamps are done now (i.e. millis before and after and nanos for duration), anyone remembers ? @vhotspur ?

vhotspur:
> As far as the various timestamps go, see #107 ? I don't know the exact reason for the way the timestamps are done now (i.e. millis before and after and nanos for duration), anyone remembers ? @vhotspur ?

As far as I remember, the reasoning was following: `nanoTime` is used to measure the duration of the operation (iteration?) as it is the most precise time source available. The unix timestamps around should have smaller overhead and they are supposed to give (1) rough duration of the whole operation including setup, (2) rough estimate of time between operations and (3) can be used to correlate data from other sources that also use wall-clock time.

lbulej:
> Code changes LGTM, however policy wise I am not sure we want the separation of before and after.
> What is the rationale to have both ?
> It makes it a bit more complex I think then it needs to be, but maybe I am missing a use-case in my head :-)

Good questions. That's what I was struggling with as well :-)

Now that I'm looking into our modified ScalaBench code, we have added one option to force GC *before operation* and another option to force GC *after benchmark*. The reason for the first is obvious. According to comments in code, the second was added to make GC stats reflect benchmark allocation more accurately.

So forcing GC after operation indeed seems to serve no purpose, unlike forcing GC after benchmark. I just did not remember the rationale for our modifications to ScalaBench properly and only recalled that we had it in two places, separately controlled.

Anyway, assuming that I fix that, do we want/need them separately controlled? 

> Regarding printing I would vote for keeping the printing that way everybody sees that gc is invoked, makes it more explicit.

Noted.

ceresek:
I think the proposal in #107 meets the reasoning above better than the current state. Also, I think some of the design was influenced by the CSV output, which does not easily permit non rectangular data.

What I think we should do with the timers:
- We should measure all things related to individual repetitions with `nanoTime` as the most accurate source, and use deltas from the beginning of repetition to make the timestamps reasonably small. Obviously this includes the "useful operation time" but possibly also "setup time" and "teardown time" ?
- We should include some way of corelating the repetition times with wall clock time that may be used in external logs, hence the `currentTimeMillis` stamp which I think we only need once per repetition, not twice.

davleopo:
@ceresek re before/after ah yes, that makes sense :-) . 

@lbulej re separately controlled gc mode: assuming no body has a use case in mind requiring that, I would vote for simplicity and only have 2 modes:  gc and no gc.

ceresek:
A bit broader comment on the issues of when to do GC (and if at all):

- Each benchmark repetition carries hidden cost of leaving garbage on the heap. In scenarios where individual repetition does not fill the heap enough to trigger GC (ideally multiple times), repetition time does not include this hidden cost.

- Triggering GC has side effects, among those I can think of right now it is tenuring all young data, and possibly cleaning various weak+soft+phantom referenced objects and putting them on reference queues.

- The cost of GC is different depending on when it is triggered. Typically, just before or just after a repetition, most data is dead and GC may be faster.

So by not doing any GC, we risk doing what we saw with the Spark benchmarks - some repetitions of the benchmark get hit by the need to do GC that accumulated over earlier repetitions, and if we do not fill memory, we may not see that work done ever. (So, if we for example compare two configurations, one where some magical optimization manages to put most temporary data on stack, and one where most temporary data is on heap, and other things remain equal, our benchmark might not see the difference. I know this example is a bit contrived.)

By forcing GC between repetitions, we avoid the risk above, but introduce other issues - one is the difference of GC cost when most data is dead, so even if we include GC time in repetition time, we may not get useful information. We may also trigger asynchronous work on the reference queue threads, introducing systematic disruption into each repetition. And if we do GC between setup and execution, we tenure all setup data, possibly increasing use of card tables or other cross generation reference tracking mechanisms that may be in place.

We may also invoke GC not to see its time cost (which is difficult to characterize anyway), but to see its effects on memory usage (effectively accounting memory usage per benchmark repetition). I'm not really sure if such accounting should include the setup and teardown allocations.

PR #197, opened 2019-08-26
Avoid unnecessary disk IO and add hash-based validation in page-rank

lbulej:
This was triggered by @davleopo mentioning leftover code in `page-rank` in the API cleanup PR :-)

### Unnecessary disk IO

When making sure that it was indeed just leftover code, I realized that maybe the setup code is causing more disk IO than necessary, so I modified the code to get rid of it.

The code originally read the whole zipped file as a single String into memory, optionally reduced the number of lines (for the `test` configuration), and wrote it uncompressed to disk. Then it used the the `SparkContext.textFile()` method to create an RDD which was then processed into a map and cached.

This uses a lazy `Source` to read either all (or a subset of) lines from the zipped file into a collection, and uses the `SparkContext.parallelize()` method to create an RDD which is then processed the same way as above. It also avoids writing the ranks from the last executed operation to disk.

This step (creating an RDD) may be something to consider, because `SparkContext.textFile()` instantiates a different implementation (Hadoop-related) of `RDD` than `SparkContext.parallelize()`. I think that in a distributed scenario, the `SparkContext.textFile()` would cause the workers to read the input file on their own from a shared file system, while the new approach reads the file in the driver (once) and then distributes the RDD. 

In our use case, this might not be a bad thing, but it should probably be an explicit decision. The measured operation execution times do not appear to be influenced (I would be surprised if they were) even though the ranks were slightly (8th decimal place after floating point) different.

I think that avoiding unnecessary disk IO is generally a good thing if it can be done, so if we are fine with this, I think we can make the other Spark benchmarks follow suit.

### Validation with floating-point ranks

Because I already had my hands on the benchmark, I also added a hash-based validation of the results in addition to checking the expected number of ranks. Even though on my machine each operation produced *identical* results, the results produced on the Travis CI machines were slightly different. Because numerical instability across different machines is a real issue to deal with, the web sites are pre-sorted sorted by ranks. This ordering is then relaxed by making neighboring ranks that differ by only a small amount equivalent, and finally the web sites are ordered by web site id within the equivalence classes.

In addition to whatever systems Travis CI uses, the default configuration passed validation on KabyLake (i7-7700HQ, 32G RAM), SkyLake (i7-6700K, 32G RAM), Sandy Bridge Xeon (E3-1245, 16G RAM), Apollo Lake Celeron (J3455, 8G RAM, OpenJDK 11), Core 2 Quad (Q9550, 4G RAM), and AMD Barcelona (Opteron 2356, 16G RAM) systems running Linux and various builds of OpenJDK 8 (mostly).

### Web site identifiers as integers

Finally, instead of keeping the web graph as `RDD[String, Iterable[String]]`, I modified the code to keep it as `RDD[Int, Iterable[Int]]` (the strings were actually numbers anyway). This seems to consume less memory and make key lookups faster. As a result, where the first operation initially took around 30 seconds on my laptop, it now takes a bit below 13 seconds, and where the operation stabilized somewhere above 3 seconds by the 19th repetition, it now stabilizes around 2.6 seconds on JDK8 and around 2 seconds on Graal 1.0.0-rc6.

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
@@ -138,10 +138,10 @@ final class PageRank extends Benchmark with SparkUtil {
 
         val rankLines = ranks
           .sortByKey()
+          .collect
           .map {
             case (url, rank) => f"$url $rank%.8f"
           }
-          .collect
           .toList
 
         Validators.hashing(expectedRankHashParam, rankLines.asJava).validate()
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
@@ -140,7 +140,7 @@ final class PageRank extends Benchmark with SparkUtil {
           .sortByKey()
           .collect
           .map {
-            case (url, rank) => f"$url $rank%.8f"
+            case (url, rank) => f"$url $rank%.7f"
           }
           .toList
 
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
@@ -1,18 +1,19 @@
 package org.renaissance.apache.spark
 
-import java.nio.charset.StandardCharsets
 import java.nio.file.Path
-import java.nio.file.Paths
+import java.util.regex.Pattern
 
-import org.apache.commons.io.FileUtils
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
+import org.renaissance.BenchmarkResult.Assert
 import org.renaissance.BenchmarkResult.Validators
 import org.renaissance.License
+
+import scala.collection.JavaConverters._
 
 @Name("page-rank")
 @Group("apache-spark")
@@ -24,14 +25,19 @@ import org.renaissance.License
   Array(
     new Parameter(name = "thread_count", defaultValue = "$cpu.count"),
     new Parameter(name = "input_line_count", defaultValue = "-1"),
-    new Parameter(name = "expected_rank_count", defaultValue = "598652")
+    new Parameter(name = "expected_rank_count", defaultValue = "598652"),
+    new Parameter(name = "expected_rank_hash", defaultValue = "d8bf7171feef1b87")
   )
 )
 @Configurations(
   Array(
     new Configuration(
       name = "test",
-      settings = Array("input_line_count = 5000", "expected_rank_count = 1661")
+      settings = Array(
+        "input_line_count = 5000",
+        "expected_rank_count = 1661",
+        "expected_rank_hash=7ca80582125b8ca7"
+      )
     ),
     new Configuration(name = "jmh")
   )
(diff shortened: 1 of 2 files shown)


vhotspur:
> ### Validation
> 
> Because I already had my hands on the benchmark, I also added a simple hash-based validation of the results (in addition to checking the expected number of ranks). Even though on my machine each operation produced _identical_ results, they were slightly different from the results produced on the Travis CI machines. Because of this, the ranks are ordered web site id and rounded to 8 decimal places to account for the numerical instability.

Would you please post here the pre-hashed data for `test` here (e.g. with the following patch)?

```diff
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
@@ -141,6 +141,9 @@ final class PageRank extends Benchmark with SparkUtil {
           .map {
             case (url, rank) => f"$url $rank%.8f"
           }
+          .map {
+            f => { println(f); f }
+          }
           .collect
           .toList
 

```

Because on my machine I got the following error for `test`:

```text
object hash: expected 7ca80582125b8ca7 but got eb987bd2440cde25
```

Tested with Java 8 and 12. I have a clean clone for this so I have no idea why I see a different result...

lbulej:
Here it is: [pre-hashed.txt](https://github.com/renaissance-benchmarks/renaissance/files/3540997/pre-hashed.txt)

vhotspur:
> Here it is: [pre-hashed.txt](https://github.com/renaissance-benchmarks/renaissance/files/3540997/pre-hashed.txt)

Thanks, this helped. The issue was caused by a different locale that I have on my laptop. Opened #198 to fix that. (With C locale, the validation is okay on my machine too.)

lbulej:
Funny that I was fixing that earlier in the printing of the execution duration (thinking that sometimes people do just parse the output and suddenly there were floating point numbers with a comma). Totally forgot about it in the validation.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
```diff
@@ -24,14 +25,19 @@ import org.renaissance.License
...
 )
 @Configurations(
   Array(
     new Configuration(
       name = "test",
-      settings = Array("input_line_count = 5000", "expected_rank_count = 1661")
+      settings = Array(
+        "input_line_count = 5000",
+        "expected_rank_count = 1661",
+        "expected_rank_hash=7ca80582125b8ca7"
```
vhotspur: Missing spaces ;-)

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/ZipResourceUtil.scala
```diff
@@ -1,21 +1,40 @@
...
 
 object ZipResourceUtil {
 
-  def readZipFromResourceToText(resourceName: String): String = {
-    val zis = new ZipInputStream(this.getClass.getResourceAsStream("/" + resourceName))
+  def sourceFromZipResource(entryName: String, resourceName: String): BufferedSource = {
+    val is = this.getClass.getResourceAsStream("/" + resourceName)
+    if (is == null) {
+      throw new FileNotFoundException(
+        s"resource '$resourceName' not found"
```
vhotspur: Perhaps prepend `/` here as well? Or the caller should take care of specifying absolute path.

farquet:
After a rebase to include Vojtech's PR, the `test` config passes, but the `default` one fails on my machine with : 
```
Benchmark 'page-rank' failed result validation:
object hash: expected d8bf7171feef1b87 but got 2e820a5be1f29c3c
```
Here is the raw data using the patch above : 
[pre-hashed-Francois.txt](https://github.com/renaissance-benchmarks/renaissance/files/3546145/pre-hashed-Francois.txt)

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/PageRank.scala
lbulej: Yikes! Fixed.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/ZipResourceUtil.scala
lbulej: You are right. I modified the utility method to require the user to pass the absolute path, it makes it more clear what is requested at the caller side.

lbulej:
I should probably learn Spark. Apparently, if you call `sortByKey()` on an `RDD` and then call `map()` on the result, you can lose the ordering. So what I do now is call `collect()` right after `sortByKey()` to get the data out of the `RDD` and then do the conversion to string outside Spark. 

The data from @farquet actually had two differences compared to mine. First was the ordering (and I discovered that even on my machine the full data were not properly sorted) and the other were differences in the last decimal place. So I changed the rounding to 7 decimal places (and updated the precomputed hashes).

Please retest if you can.

farquet:
Test config works as expected, but the main one doesn't :
```
====== page-rank (apache-spark) [default], iteration 0 started ======
Benchmark 'page-rank' failed result validation:
object hash: expected 4704db55b9f60104 but got 92d7e79d5544cdf7
The following benchmarks failed: page-rank
```

lbulej:
Despite the tiny changes, Travis fails on me, so I'm checking what results are produced on Travis. We might need to do much more rounding than expected. @farquet can you again upload the results for the default configuration?

farquet:
Hashes to `92d7e79d5544cdf7`.
[pre-hashed-v2.txt](https://github.com/renaissance-benchmarks/renaissance/files/3551247/pre-hashed-v2.txt)

lbulej:
OK, one more try with different approach (the commit message contains some details) :-) 
After chasing the wrong target for a while, I think it should be less fragile.

One more thing -- instead of keeping the web graph as RDD[String, Iterable[String]], I modified the code to keep it as RDD[Int, Iterable[Int]] (the strings were actually numbers anyway). This seems to consume less memory and make key lookups faster. As a result, where the first operation initially took around 30 seconds on my laptop, it now takes a bit below 13 seconds, and where the operation stabilized somewhere above 3 seconds by the 19th repetition, it now stabilizes around 2.6 seconds on JDK8 and around 2 seconds on Graal 1.0.0-rc6.

Is that a big problem so that I should switch back to using strings for keys?

lbulej:
Good to hear, Francois. In addition to whatever Travis CI uses, I have also tested it on KabyLake (i7-7700HQ, 32G RAM), SkyLake (i7-6700K, 32G RAM), Sandy Bridge Xeon (E3-1245, 16G RAM), Apollo Lake Celeron (J3455, 8G RAM, OpenJDK 11), Core 2 Quad (Q9550, 4G RAM), and AMD Barcelona (Opteron 2356, 16G RAM) systems running Linux and various builds of OpenJDK 8 (mostly). AMD iron (and silicon) was decidedly in short supply.

However, the validation did pass on those systems, so I would consider it reasonably stable and merge it (unless somebody screams loudly).

Issue #185, opened 2019-07-15
Use SingleShotTime in JMH wrappers


axel22:
http://javadox.com/org.openjdk.jmh/jmh-core/0.8/org/openjdk/jmh/annotations/Mode.html

This would be better than using the 1 ms default time, which is currently being emitted.

PR #201, opened 2019-09-13
Issue #185 - Use SingleShotTime in JMH wrappers

ericcaspole:
Switch  @BenchmarkMode to Mode.SingleShotTime and remove unneeded timing parameters from @Warmup and @Measurement as a result of that.


axel22:
LGTM - thank you!

PR #204, opened 2019-10-16
Introduce a new CSS class for the rightmost li element of the website.

axel22:
This is a part of the solution to the current website issue.


axel22:
See also: https://github.com/renaissance-benchmarks/renaissance/pull/205

PR #205, opened 2019-10-16
Add the new CSS class to the layout

axel22:
This is the second part of the solution for the website issue.

Before:

screenshot-17-25-31

After:

screenshot-17-25-25


axel22:
We will still need to do redeployment at some point.

farquet:
Go for it ;)

Issue #83, opened 2019-04-28
Better error handling in neo4j-analytics


ceresek:
When terminating neo4j-analytics by killing its threads (necessary when the harness does not have support for execution timeout), the benchmark can reach a state where the database is shut down, accesses report `org.neo4j.graphdb.DatabaseShutdownException: This database is shutdown.` but the benchmark happily reports nonsensical repetition times (from 5-6 seconds per repetition to sub ms per repetition).

The benchmark should not report repetition times when the repetitions are obviously failing. We should check this for all benchmarks (that is, include result validation and report at least a warning if the results do not validate).

[neo4j-analytics-stderr.log](https://github.com/D-iii-S/renaissance-benchmarks/files/3124704/neo4j-analytics-stderr.log)
[neo4j-analytics-stdout.log](https://github.com/D-iii-S/renaissance-benchmarks/files/3124705/neo4j-analytics-stdout.log)

farquet:
Fixed in #224.

PR #212, opened 2019-12-17
Updates for Neo4j Benchmark

jexp:
- added index/constraints
- wait for constraints to be online
- improved some of the cypher statements
- switched to parameter usage
- 
- upgrade to Neo4j 3.5.12
- configure page-cache
- tried to add result validation, but I don't think it uses that

### DIFF
--- a/benchmarks/neo4j/build.sbt
+++ b/benchmarks/neo4j/build.sbt
@@ -6,10 +6,11 @@ lazy val neo4j = (project in file("."))
     version := (version in renaissanceCore).value,
     organization := (organization in renaissanceCore).value,
     scalafmtConfig := Some(file(".scalafmt.conf")),
-    scalaVersion := "2.11.4",
+    scalaVersion := "2.11.12",
+    scalacOptions += "-target:jvm-1.8",
     libraryDependencies ++= Seq(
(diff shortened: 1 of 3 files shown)


farquet:
Thanks a lot @jexp for the PR ! This is very welcome to have your Neo4J expertise on this benchmark.

Sorry that the review took so long. I'm currently extending your work with integration into our validation mechanism while also fixing #83 where interruption of the benchmark was not properly captured.

I'll send a patch here soon. If you could apply it to the PR, that would be appreciated.

farquet:
Hi Michael,
Could you apply that patch to your branch please ?
That fixes the validation and ensure that the benchmark fails properly when the benchmark is interrupted.
Also note that it fixes the checkstyle (this is the reason of the CI failure). The code can be automatically formatted with scalafmt by calling `tools/sbt/bin/sbt renaissanceFormat`.
Thanks !

[0001-Validation-of-neo4j-analytics-benchmark.patch.txt](https://github.com/renaissance-benchmarks/renaissance/files/4102501/0001-Validation-of-neo4j-analytics-benchmark.patch.txt)

farquet:
Merged in #224.
Thanks again @jexp for the contribution !

Issue #163, opened 2019-06-24
db-shootout crashes with jdk-12


albertnetymk:
Downloaded `jdk-12.0.1+12` from https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=hotspot#x64_linux and renaissance jar from the release page.

Encountered the following crash with `java -jar renaissance-mit-0.9.0.jar -r 1 db-shootout`:

```
Error during tear-down: null
java.lang.NullPointerException
        at org.lmdbjava.bench.Chronicle$CommonChronicleMap.teardown(Chronicle.java:131)
        at org.lmdbjava.bench.Chronicle$Reader.teardown(Chronicle.java:210)
        at org.renaissance.database.DbShootout.tearDownAfterAll(DbShootout.scala:81)
        at org.renaissance.RenaissanceBenchmark.runBenchmark(RenaissanceBenchmark.java:97)
        at org.renaissance.RenaissanceSuite$.$anonfun$main$2(renaissance-suite.scala:308)
        at org.renaissance.RenaissanceSuite$.$anonfun$main$2$adapted(renaissance-suite.scala:306)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
        at org.renaissance.RenaissanceSuite$.main(renaissance-suite.scala:306)
        at org.renaissance.RenaissanceSuite.main(renaissance-suite.scala)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:567)
        at org.renaissance.Launcher.main(Launcher.java:18)
Exception occurred in org.renaissance.database.DbShootout@58ce9668: java.lang.NoSuchFieldException: override
java.lang.AssertionError: java.lang.NoSuchFieldException: override
        at net.openhft.chronicle.core.Jvm.setAccessible(Jvm.java:326)
[...]

PR #213, opened 2020-01-07
[WiP] Test with newer JDKs on Travis

vhotspur:
Opening to track progress, current issues:

 * [ ] #163 (db-shootout on JDK 12)
 * [ ] Failing builds with JDK 13 (update of SBT is [probably needed](https://github.com/sbt/sbt/issues/5093))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ jobs:
 
     - <<: *bundle
       os: osx
+      osx_image: xcode9.3
+      jdk: openjdk8
     - <<: *bundle
       jdk: openjdk8
     - <<: *bundle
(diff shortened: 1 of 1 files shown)


RobertHenry6bev:
Has there been any progress on getting this to build out-of-the-box with a version of Java > 11?  We'd like to use the latest Java 14

ceresek:
Did a quick check to see what needs to be done, using OpenJDK 14.0.1.

Looks like this will need updating Scala from 2.12.8 to at least 2.12.11 (https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html) and sbt to who-knows-what version (currently recommended on that same page is 1.3.9 or newer, I used 1.3.13).

This in turn breaks compilation for dotty, the rest of the benchmarks compiles without fatal errors. In the resulting archive, db-shootout fails, the rest of the benchmarks can complete at least one iteration.

vhotspur:
Closing, superseded by #271.

PR #216, opened 2020-01-21
Add VM uptime timestamps to repetitions and (much) more environmental info

lbulej:
This adds VM uptime timestamps (in nanoseconds) to the start of measured operations, as discussed in #107. To determine the value of `nanoTime()` at VM start, we use the average of two `nanoTime()` readings straddling two distinct but consecutive `currentTimeMillis()` values so that we get somewhere in the middle of the platforms `currentTimeMillis()` accuracy. 

Because this changes the version of the JSON output format, I also bundled changes which add much more info about the environment to the JSON output file (basically everything there is that you get in the `VM Summary` tab of `jconsole`). While the information is not going to be used most of the time, it does provide a more colorful picture of the VM the results were collected on (which GC, which compilation system, etc) which is sometimes useful (in retrospect).

All commits except the last massage the `ResultWriter` so that the very last commit (which actually adds the timestamps and the extra info) looks simple :-)

### DIFF
--- a/renaissance-harness/src/main/scala/org/renaissance/harness/RenaissanceSuite.scala
+++ b/renaissance-harness/src/main/scala/org/renaissance/harness/RenaissanceSuite.scala
@@ -150,35 +150,57 @@ object RenaissanceSuite {
 
   private def getVmStartNanos = {
     //
-    // Get nanoTime() reading from between two
-    // distinct currentTimeMillis() readings.
+    // Get two nanoTime() samples around currentTimeMillis() that are as
+    // close as possible to when the currentTimeMillis() result flips. Use
+    // the two straddling nanoTime() readings as an estimate of nanoTime()
+    // corresponding to the last currentTimeMillis() result.
     //
-    val initMillis = System.currentTimeMillis()
-    while (System.currentTimeMillis() == initMillis) {
-      // Wait for the next reading.
-    }
+    // Repeat to minimize the difference between the two nanoTime() readings
+    // and stop when no improvement has been observed for several iterations.
+    //
+    // This avoids basing the estimate on first calls to native methods which
+    // tend to be slow, and has a chance to avoid scheduling artifacts.
+    //
+    val streakLengthMax = 10
 
-    // Wait for the next reading and get straddling nanoTime readings.
-    val currentMillis = System.currentTimeMillis()
-    var currentNanosBefore = System.nanoTime ()
-    var currentNanosAfter = 0L
-    var sameTick = true
-    while (sameTick) {
-      sameTick = (System.currentTimeMillis() == currentMillis)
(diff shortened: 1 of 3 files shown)


ceresek:
Fails with JDK 11:
```
SEVERE: Harness failed with exception: java.lang.UnsupportedOperationException: Boot class path mechanism is not supported
	at java.management/sun.management.RuntimeImpl.getBootClassPath(RuntimeImpl.java:99)
	at org.renaissance.harness.JsonWriter.getJreInfo(ResultWriter.scala:222)
...
```

renaissance-harness/src/main/scala/org/renaissance/harness/RenaissanceSuite.scala
```diff
@@ -145,6 +148,35 @@ object RenaissanceSuite {
...
+    //
+    // Get nanoTime() reading from between two
+    // distinct currentTimeMillis() readings.
+    //
+    val initMillis = System.currentTimeMillis()
+    while (System.currentTimeMillis() == initMillis) {
+      // Wait for the next reading.
+    }
+
+    val currentNanosBefore = System.nanoTime()
```
ceresek: Don't you want to make it so that `currentNanosBefore` is the last timestamp taken before millis change, not the first one ?

lbulej:
Hmm, the builds seem to fail randomly due to unresolved dependencies:

```
[error] sbt.librarymanagement.ResolveException: unresolved dependency: org.eclipse.collections#eclipse-collections-api;[7.0.0,7.20.0): not found
[error] unresolved dependency: org.eclipse.collections#eclipse-collections;[7.0.0,7.20.0): not found
[error] unresolved dependency: org.eclipse.collections#eclipse-collections-forkjoin;[7.0.0,7.20.0): not found
[error] unresolved dependency: com.google.guava#guava;[15.0,19.20): not found
```

What baffles me that I no longer (actually, for some time already) have the option to restart the builds on the Travis website.

vhotspur:
> What baffles me that I no longer (actually, for some time already) have the option to restart the builds on the Travis website.

I restarted the builds, they pass now. I double-checked your permissions and I have not found anything wrong. I also looked through some Travis tickets and you are not the only one with similar issues. For some, resyncing their Travis account and/or re-logging-in helped. If that does not help, I would say you ought to contact Travis as you are repo owner here and you shall be able to restart/cancel any job...

lbulej:
> > What baffles me that I no longer (actually, for some time already) have the option to restart the builds on the Travis website.
> 
> I restarted the builds, they pass now. I double-checked your permissions and I have not found anything wrong. I also looked through some Travis tickets and you are not the only one with similar issues. For some, resyncing their Travis account and/or re-logging-in helped.

Right. Re-logging-in did help. Silly I did not think about it earlier.

Issue #165, opened 2019-06-28
Core JAR included multiple times


axel22:
Reported by @tkrodriguez : we have some of the dependencies JARs multiplied across different subprojects. This inflates the final JAR size a bit, but I think it's not an immediate cause for alarm.

```
2 leveldbjni-all-1.8.jar
2 aopalliance-1.0.jar
2 scala-reflect-2.11.8.jar
2 opencsv-2.3.jar
2 lz4-1.3.0.jar
2 paranamer-2.8.jar
2 javax.inject-1.jar
2 commons-compress-1.12.jar
2 scala-parser-combinators_2.11-1.0.4.jar
2 guava-19.0.jar
5 scala-library-2.12.8.jar
5 scala-library-2.11.8.jar
6 commons-io-2.6.jar
13 renaissance-core-0.9.0.jar 
```

Long-term, we should probably implement some sort of compression for JAR duplicates.

A more immediate action might be to figure out why we duplicate the `renaissance-core` library across projects. @vhotspur @lbulej Were you aware of this? Seems like we're including the core unnecessarily in the subproject JARs:

https://github.com/renaissance-benchmarks/renaissance/blob/master/build.sbt#L50

lbulej:
As of #267 the `renaissance-core` library is considered a `provided` dependency and is not packaged at all, because its classes are provided directly by the top-level bundle. This does nothing about scala library classes though, but I'm closing this for now.

Issue #137, opened 2019-05-15
Add sbt command to clean all class files


farquet:
I noticed that running `clean` in sbt, even followed by `cleanFiles` keeps a lot of class files around. If you are switching between JDKs, you will hit bad class file versions and you will have to remove them manually somehow.
Currently, I simply do 
```
find . -type d -name 'target' -prune -print -exec rm -rf {} ';'
```

but that's dangerous and removes more than necessary.
We should introduce an sbt command `renaissanceClean` that properly removes those class files.

Reference : https://stackoverflow.com/questions/4483230/an-easy-way-to-get-rid-of-everything-generated-by-sbt

vhotspur:
I found this [old issue in SBT tracker](https://github.com/sbt/sbt/issues/896#issuecomment-25752887) where it seems that it is actually not an issue at all ;-)

> [...] been discussed on the mailing list before with the same conclusion as the linked StackOverflow question. If you really want a clean cache, [...] you don't need to rely on the build tool's clean. [...]

PR #221, opened 2020-02-04
Ant build system experiments

### LINKED ISSUE #186: JMH version doesnt build some time after rev e895d544ee7ddbd14caf975072fd51717224b509
Building the JMH version today gives an error:
```
$ tools/sbt/bin/sbt clean renaissanceJmh/jmh:assembly
...

[error] Caused by: java.lang.ClassNotFoundException: org.renaissance.dummy.Jmh_Dummy
[error] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
[error] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
...
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] (renaissanceJmh / Jmh / generateJmhSourcesAndResources) java.lang.NoClassDefFoundError: org/renaissance/dummy/Jmh_Dummy
[error] Total time: 203 s, completed Jul 15, 2019 4:01:09 PM
```

@tkrodriguez mentioned he also sees this error while building.
Thanks,
Eric

vhotspur:
This PR is meant to collect opinions and thoughts whether we shall switch to Ant build system instead of SBT.

# The ``Why''

My first reason to experiment with Ant was the complexity of SBT. While it is quite nice that build scripts for individual benchmarks are rather short, the complexity of the main script is quite terrifying and requires a deep knowledge of many SBT concepts.

Personally I also do not like the fact that SBT hides so many things. For example, did you know that SBT silently creates emulated `rt.jar` on JDK9+ where this JAR was replaced by modules?

The speed of the build is the second major issue of SBT. Our Travis jobs currently take about 8 minutes to build the suite. Multiplied by the number of different JDKs, it means that the (compilation) results are available after about 20 minutes.

But even the incremental build is quite slow (and I ignore here the start-up cost of launching SBT shell). And changing anything that requires project reload is often even worse than starting the shell anew.

# Migration to Ant

Because of these shortcomings (at least from my point of view) I tried to convert the build system to Ant and Ivy. Just to try how it would look like, whether it is possible and what would be the speed improvement (if any).

The work is not complete yet (see missing things at the bottom) but has the major parts to allow us to evaluate whether to switch or not.

The Ant scripts are currently able to compile all benchmarks and they use Ivy to manage the dependencies. The build creates the fat JAR with all dependencies that executes on JDK 8, 9 and 11 on Travis (plus JDK12 build without execution because of #163).

The truth is that the new scripts are about the same size as SBT ones but I really believe they are simpler as they are more explicit. Every step is explicitly covered by specific tasks, there are no hidden
dependencies or anything that happens automagically.

On my machine, the build is almost twice as fast (both for clean and incremental builds) and I have not yet experimented with the `<parallel>` task in Ant. The bottleneck is usually the packaging of the final JAR that alone takes about 20 seconds.

# How To Run

If you want to try it out, check-out branch `topic/ant-experiments` and simply run `ant`. That would create `out/renaissance-0.10.0.jar` with the whole suite.

It is also possible to build JAR with one benchmark group only via `ant one -Done.name=jdk-streams` that produces `out/one-jdk-streams.jar` that will contain only benchmarks from `jdk-streams` group.
That is useful when debugging one benchmark only.

Running `ant clean` removes *all* generated files (`ant ivy-purge` removes the whole Ivy cache that usually resides in `~/.ivy2/cache`).

# Missing Features

Currently missing features are the following ones. I will add them only if we decide to go ahead and switch the build system as they are IMHO not crucial for trying things out.

* [ ] Packaging of JMH.
* [ ] MIT and GPL builds.
* [ ] Service targets for push hooks etc.
* [ ] A bit of refactoring (target names unifying).

As a nice side effect, this would also allow us to close #186, #165 and #137. And with the reduced compilation time, we might be able to test regularly on [AppVeyor](https://www.appveyor.com/about/) where the time-outs are shorter than on Travis.


axel22:
Some feedback from sbt people:

```
@axel22 the claims are pretty wild, but either way I'd suggest updating to sbt 1.3 which has faster dep management via coursierand perhaps using nativepackager instead of assembly
also not sure why you build with different jdks, the resulting jars should be identical so you can build just once
```

vhotspur:
> Some feedback from sbt people:

Isn't this something that is called conflict of interest? :-D

> the claims are pretty wild, but either way I'd suggest updating to sbt 1.3 which has faster dep management via coursier and perhaps using nativepackager instead of assembly

I do not think that you can actually speed-up the JAR generation as packaging 400M takes some time :-)

As a matter of fact, I tried (though quickly) updating SBT and something broke without any direct hint of what went wrong. And the manual was not really helpful.

> also not sure why you build with different jdks, the resulting jars should be identical so you can build just once

Not really. The class versions are different so you cannot build only with the newest. But it would be weird if we would not be able to compile on new JDKs. So I think we need to build on as many JDKs as possible.

axel22:
> > Some feedback from sbt people:
> 
> Isn't this something that is called conflict of interest? :-D
> 

Well, they're certainly on the sbt band wagon, but I thought that they might have some information that we don't.

> > the claims are pretty wild, but either way I'd suggest updating to sbt 1.3 which has faster dep management via coursier and perhaps using nativepackager instead of assembly
> 
> I do not think that you can actually speed-up the JAR generation as packaging 400M takes some time :-)

The way I understood it, he's suggesting that nativepackager is faster than the sbt-assembly plugin, and that the initial dependency management that causes a slow sbt startup could be made faster with coursier (that was taken from the gitter chat at sbt/sbt).

> As a matter of fact, I tried (though quickly) updating SBT and something broke without any direct hint of what went wrong. And the manual was not really helpful.
> 
> > also not sure why you build with different jdks, the resulting jars should be identical so you can build just once
> 
> Not really. The class versions are different so you cannot build only with the newest. But it would be weird if we would not be able to compile on new JDKs. So I think we need to build on as many JDKs as possible.

vhotspur:
> > > the claims are pretty wild, but either way I'd suggest updating to sbt 1.3 which has faster dep management via coursier and perhaps using nativepackager instead of assembly
> > 
> > 
> > I do not think that you can actually speed-up the JAR generation as packaging 400M takes some time :-)
> 
> The way I understood it, he's suggesting that nativepackager is faster than the sbt-assembly plugin, and that the initial dependency management that causes a slow sbt startup could be made faster with coursier (that was taken from the gitter chat at sbt/sbt).

Might be, have not tried. I have seen coursier some time ago and it definitely was not stable enough to be tried. Could be better now.

But for me the really annoying thing is the inability to clean files reliably. Perhaps that is also better with new versions?

axel22:
To summarize some of the things we discussed offline: I'm in principle OK with an Ant rewrite in case the rewrite reproduces all the functionality that we already have in SBT, and in particular if you want to maintain the new build.

I would, however, suggest that you list a precise descriptions of the problems that we have with the current SBT build, and that you share that precise list with the SBT community (gitter and issues), as this would confirm that there are no easy workarounds for those problems (because there might be obvious things that we are missing, because we don't know enough about SBT).

vhotspur:
Closing. Some of the things were solved by #267 and #271 and Ant is now even less supported by Scala community.

PR #225, opened 2020-03-22
Add JDK8 and JDK11 charts on the main page

farquet:
Website update to show the latest numbers from https://github.com/renaissance-benchmarks/measurements/pull/4/

Each JDK is shown in a different tab, making it future proof.

This assumes that, once the above PR will be merged, the JDK 8 and JDK 11 charts will be available at location : 

- https://github.com/renaissance-benchmarks/measurements/raw/master/stripe-jdk-8.png
- https://github.com/renaissance-benchmarks/measurements/raw/master/stripe-jdk-11.png

JDK8
JDK11

### DIFF
--- a/website/_layouts/mainrenaissance.html
+++ b/website/_layouts/mainrenaissance.html
@@ -144,7 +144,14 @@ layout: base
   <div class="container_16">
     <div class="grid_16">
       <div id="measurements" class="text-snippet">
-        <img src="https://github.com/renaissance-benchmarks/measurements/raw/master/stripe.png"/>
+        <ul id="jdk-tabs">
+
+          <li><a id="jdk8">JDK 8</a></li>
+          <li><a id="jdk11">JDK 11</a></li>
+
+        </ul>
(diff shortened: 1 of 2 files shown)


website/resources/styles/proj.css
```diff
@@ -1139,3 +1139,94 @@ a.faint-link:hover {
 #measurements img {
   width: 960px;
 }
+
+#jdk-tabs {
```
axel22: Opt: you could use a class instead of an ID.

website/_layouts/mainrenaissance.html
```diff
@@ -144,14 +144,41 @@
...
   </div>
+  <script type='text/javascript'>
+  $(document).ready(function() {
+
+  $('#jdk-tabs li a:not(:first)').addClass('inactive');
+  $('.jdk-container').hide();
+  $('.jdk-container:first').show();
+
+  $('#jdk-tabs li a').click(function(){
+      var t = $(this).attr('id');
```
axel22: 2 indents.

Issue #227, opened 2020-04-14
movie-lens reloads data in each repetition (and does not cache some RRDs)


lbulej:
I was checking other Spark-based in response to #226 to make sure they all cache their RRDs, and when I got to movie lens `movie-lens` and I found the following:

There are 5 RDDs (`ratings`, `personalRatingsRDD`, `training`, `validation`, and `test`), and we only call `cache()` on those assigned to `training`, `validation`, `test` in the `splitRatings()` method, which makes use of the `personalRatingsRDD` in the expression producing RDD for `training`. 

Neither does the RDD assigned to `ratings` in the `loadRatings()` method get cached.

I'm not sure if that is an actual problem, but both `splitRatings()` and `loadRatings()` are called on each benchmark repetition. `splitRatings()` is called directly from the `run()` method, while `loadRatings()` is called from the `run()` method through the `loadData()` method, which I think is a major problem.

The `loadData()` method keeps writing to disk and reloading the data in each repetition, even though this is something that I would expect to be done in the `setUpBeforeAll()` method.

Can someone double check that I read this correctly?

farquet:
I wonder how we let that slip through :)

I agree with all your observations. Note that there are a few calls to `cache()` (on `training`, `validation` and `test`), but they won't prevent recomputation since they are part of the core of the iteration. :/

My suggestion would be to ensure caching of all 5 RDDs, instantiate the `MovieLensHelper` and call the setup/load methods only once in the `setupBeforeAll()` method.
I am hesitant in keeping the training/test/validation split part of the core of the iteration, but after some local testing, I see that the iteration time is highly dominated by the ALS training, so this isn't very important.

On top of PR #228, my suggested changes are below. Do we conceptually agree on those changes before I push them to the branch ?
Moreover, I would be curious to know if this is somewhat related to #187.

```
diff --git a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala
index 415fb4e..9d3ad19 100644
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala
@@ -72,6 +72,8 @@ final class MovieLens extends Benchmark with SparkUtil {
 
   var sc: SparkContext = _
 
[...]

PR #228, opened 2020-04-14
Address RDD caching problems

### LINKED ISSUE #226: log-regression benchmark doesn't cache its RDD
All Spark benchmarks are structured the same way and start by loading data from files, then transform it in an RDD and then run the benchmark on the cached RDD.
However, the `log-regression` benchmark doesn't cache its RDD, recomputing it during the iteration of the benchmarks. This is an inconsistency and a problem we have to fix.

Credit goes to @tkrodriguez for spotting the issue.

```
diff --git a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala
index ee5daad..0fefa91 100644
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala
@@ -96,7 +96,7 @@ final class LogRegression extends Benchmark with SparkUtil {
           features(index) = value
         }
         (parts(0).toDouble, Vectors.dense(features))
-      }
+      }.cache();
   }
 
   override def setUpBeforeAll(c: BenchmarkContext): Unit = {
```

farquet:
This fixes #226 .
Moreover, since we don't really have unit tests, I've added a `checkCaching` method to the `SparkUtil` trait. That way we can explicitly mark some RDDs that need to be cached and fail otherwise. That will make things explicit and avoid such problems in the future.
Before I start porting this extra check to all Spark benchmarks that use RDDs, please let me know if you think this idea makes sense (or if you see a better way to achieve the same thing).

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
@@ -118,7 +118,7 @@ final class DecTree extends Benchmark with SparkUtil {
 
     tempDirPath = c.generateTempDir("dec_tree")
     sc = setUpSparkContext(tempDirPath, threadCountParam, "dec-tree")
-    training = prepareAndLoadInput(decisionTreePath, inputFile)
+    training = prepareAndLoadInput(decisionTreePath, inputFile).cache()
     pipeline = constructPipeline()
   }
 
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
@@ -8,6 +8,8 @@ import java.nio.file.Paths
 import org.apache.commons.io.IOUtils
 import org.apache.spark.SparkConf
 import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+
 
 trait SparkUtil {
 
@@ -57,6 +59,12 @@ trait SparkUtil {
         new FileOutputStream(winutilsPath.toString + winUtils)
       )
       System.setProperty("hadoop.home.dir", tempDirPath.toAbsolutePath.toString)
+    }
+  }
+
+  def checkCaching[T](rdd: RDD[T]): Unit = {
+    if (!rdd.getStorageLevel.useMemory) {
+      throw new Exception("Spark RDD must be cached !")
     }
   }
 }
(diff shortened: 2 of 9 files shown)


benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
```diff
@@ -59,4 +61,10 @@ trait SparkUtil {
       System.setProperty("hadoop.home.dir", tempDirPath.toAbsolutePath.toString)
     }
   }
+
+  def checkCaching[T](rdd: RDD[T]): Unit = {
+    if (!rdd.getStorageLevel.useMemory) {
```
farquet, line 66: `rdd.getStorageLevel.useMemory` is true when `cache()` has been called on the RDD.

davleopo:
LGTM, seems to do the trick

lbulej:
LGTM.

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/DecTree.scala
```diff
@@ -118,7 +118,7 @@ final class DecTree extends Benchmark with SparkUtil {
 
     tempDirPath = c.generateTempDir("dec_tree")
     sc = setUpSparkContext(tempDirPath, threadCountParam, "dec-tree")
-    training = prepareAndLoadInput(decisionTreePath, inputFile)
+    training = prepareAndLoadInput(decisionTreePath, inputFile).cache()
```
lbulej, line 121: Good catch, I was only searching for RDD type.

farquet:
The `dec-tree` benchmark contains a spark SQL DataFrame set to `sqlContext.read.format("libsvm").load(bigInputFile.toString)`.
This is also lazy evaluated and not cached, so adding a `.cache()` at the that end of that line reduces the iteration time by ~30% (similarly on different JDKs tested).

Unfortunately, I can't find a way to check if a DataFrame is cached in the Spark documentation. So I can't have an `ensureCaching` method that takes a DataFrame as input.

farquet:
Also addresses #227 now. `movie-lens` had quite some setup (or missing caches) happening in the core of the iteration. This reduces the iteration time by 15-20% which are taken from the initial bits of the execution, the rest (als algo training) is untouched.

farquet:
I wonder why there is no Travis check running on this branch ?
I guess no check means no problem ! Merging.

lbulej:
> I wonder why there is no Travis check running on this branch ?

I wonder too! I realized that Travis did not run for my earlier PR too...

vhotspur:
> I wonder why there is no Travis check running on this branch ?
> I guess no check means no problem ! Merging.

The checks were running: https://travis-ci.org/github/renaissance-benchmarks/renaissance/pull_requests

But their status was not reported back... Might be related to https://www.traviscistatus.com/incidents/3krnmsdpp3t0

PR #229, opened 2020-04-14
Don't include dummy benchmarks/groups in output of CLI list commands

### LINKED ISSUE #220: Inclusion of dummy benchmarks in list complicates scripting
The `--raw-list` option would be great for scripting the usual "run all benchmarks" task, but it also includes the `dummy` group of benchmarks, which the script has to filter afterwards. Having an option to list the benchmarks without the artificial ones that are not meant for use would be useful.

lbulej:
Removes `dummy` benchmarks and groups from the output of the `--list`, `--raw-list`, and `--group-list` CLI commands to remove the need to filter out these in scripts. This fixes #220. 

This is consistent with the fact that dummy benchmarks are ignored when the user requests to run `all` benchmarks, where `all` is considered to be a special group containing all (public) benchmarks.

I'm not so sure if this is the right behavior for the `--list` command, which is not usually consumed by scripts, but my OCD was happier when I filtered it everywhere :-)

This does not affect the list in `README.md`, where these dummy benchmarks/groups are still listed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ refs:
       - 'test $(unzip -p renaissance-jmh/target/scala-2.12/renaissance-jmh-assembly-$RENAISSANCE_VERSION.jar META-INF/BenchmarkList | wc -l) -gt 0'
       - 'java -version'
       # Test benchmarks under Renaissance harness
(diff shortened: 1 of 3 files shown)


farquet:
I don't see the Travis CI here either. Just like #228. Is this a known problem ?

PR #242, opened 2021-04-20
Update benchmarks to use Scala 2.12 or 2.13

lbulej:
While this branch started with a simple idea of updating Scala in various benchmarks to recent versions, it turned out to be a much more involved endeavor, resulting in a lot of fallout that needed handling. To make benchmarks compatible with newer Scala, their dependencies needed to be updated, sometimes substantially (e.g., Spark, Neo4J, Dotty). However, when the dust settles, the suite should be more compatible with modern JVMs. There are a few things that need still need to be done before release, but these should come through much smaller PRs.

### Key changes

- Suite uses Scala versions 2.12.13 and 2.13.5 only, Scala 2.11 is no longer used
- Spark benchmarks now use Spark 3.0.1, Spark 2.0.0 is no longer used
- Benchmarks can be in more than one benchmark group
- Benchmarks can specify requirements on JVM versions
- Harness avoids executing benchmarks on incompatible JVM
- Harness provides better control over scratch directories
- Build system uses SBT 1.4.9 (and some updated plugins)

### Follow-up changes

- Replace the contents of `sources.zip` in `dotty` benchmark with `scalap` sources
- Clean-up scratch file and directory handling in all benchmarks
- Capture requirements on JVM versions for the rest of the benchmarks
- [x] Update `README.md` generator to output benchmark JVM version requirements
- Handle Spark parameters consistently in Spark benchmarks

### Used Scala versions

The following benchmarks now use Scala 2.12:
- `reactors`
- **apache-spark** group (`als`, `chi-square`, `dec-tree`, `gauss-mix`, `log-regression`, `movie-lens`, `naive-bayes`, `page-rank`)
- `neo4j-analytics`
- **scala-stm** group (`philosophers`, `scala-stm`)
- **twitter-finagle** group (`finagle-chirper`, `finagle-http`)

The following benchmarks use Scala 2.13:
- `akka-uct`
- `dotty`
- `scala-doku`
- `scala-kmeans`

The following are Java benchmarks with wrappers using Scala 2.13:
- `db-shootout`
- **jdk-concurrent** group (`fj-kmeans`, `future-genetic`)
- **jdk-streams** group (`mnemonics`, `par-mnemonics`, `scrabble`)
- `rx-scrabble`

### Benchmark changes

- Benchmark can be in more than one benchmark group. This can be achieved by providing the benchmark class with multiple `@Group` annotations.
- Benchmarks can specify requirements on JVM versions. There are two new annotations for this purpose: the `@JvmRequired` annotation, which allows setting the minimum required JVM version a benchmark needs to run (inclusive, defaults to `"1.8"` if unset), and the `@JvmSupported` annotation, which allows setting the maximum supported JVM version a benchmark runs on (also inclusive, unset by default).

The following provides a more detailed summary of changes in individual benchmarks.

#### actors/akka-uct (Scala 2.13)

- Updated Scala version (2.11.8 -> 2.13.5)
- Split off into separate subproject to make it independent of `reactors`
- Updated `com.typesafe.actor.akka-actor` dependency (2.3.11 -> 2.16.12)

#### actors/reactors (Scala 2.12)

- Updated Scala version (2.11.8 -> 2.12.13)
- Split off into separate subproject to make it independent of `akka-uct`
- Updated `io.reactors.reactors-core` dependency (0.7 -> 0.9-renaissance-83d194)
  - The dependency was ported/cleaned from Scala 2.11 to Scala 2.12
  - https://github.com/renaissance-benchmarks/dependency-reactors/tree/renaissance/export

#### apache-spark/* (Scala 2.12)

- Updated Scala version (2.11.8 -> 2.12.13)
  - Spark 3.0 does not support Scala 2.13
- Updated `org.apache.spark.spark-core` dependency (2.0.0 -> 3.0.1)
- Updated `org.apache.spark.spark-sql`  dependency (2.0.0 -> 3.0.1)
- Updated `org.apache.spark.spark-mllib`  dependency (2.0.0 -> 3.0.1)

#### database/db-shootout (Scala 2.13)

- Scala wrapper over Java benchmark
- Updated Scala version (2.11.8 -> 2.13.5)

#### dummy/* (Java only)

- Internal benchmarks for harness testing
- Made into pure Java subproject

#### jdk-concurrent/* (Scala 2.13)

- Scala wrapper over Java benchmarks
- Updated Scala version (2.12.8 -> 2.13.5)

#### jdk-streams/* (Scala 2.13)

- Scala wrapper over Java benchmarks
- Updated Scala version (2.12.8 -> 2.13.5)

#### neo4j/neo4j-analytics (Scala 2.12)

- Updated Scala version (2.12.8 -> 2.13.5)
- Benchmark ported to Neo4J 4.x API
- Cleaned up scratch file handling
- Updated `net.liftweb.lift-json` dependency (3.2.0 -> 3.4.3)
- Updated `org.neo4j.neo4j` dependency (3.5.12 -> 4.2.4)
  - Neo4J 4.2 does not support Scala 2.13
- Removed direct `commons-io.commons-io` dependency

#### rx/rx-scrabble (Scala 2.13)

- Scala wrapper over Java benchmark
- Updated Scala version (2.11.8 -> 2.13.5)

#### scala-dotty/dotty (Scala 2.13)

- Updated Scala version (2.12.8 -> 2.13.5)
- Updated contents of `sources.zip` to make it compatible with updated Dotty
  - The plan is to replace the contents with `scalap` sources
- Added `org.scala-lang.modules.scala-collection-compat` dependency (2.4.2)
- Added `org.scala-lang.scala3-compiler_3.0.0-RC1` dependency (3.0.0-RC1)
- Removed direct `ch.epfl.lamp.dotty-compiler_0.12` dependency
- Removed direct `commons-io.commons-io` dependency

#### scala-sat/scala-doku (Scala 2.13)

- Updated Scala version (2.11.7 -> 2.13.5)
- Updated bundled dependency on CafeSat (0.01 -> 0.01-28-gd0edeaa)
  - The dependency was ported/cleaned from Scala 2.11 to Scala 2.13
  - https://github.com/renaissance-benchmarks/dependency-cafesat/tree/renaissance/export
- Updated bundled dependency on `com.regblanc.scala-smtlib` (0.1 -> 0.2.1-52-ga71d6b0)
  - The dependency was ported/cleaned from Scala 2.11 to Scala 2.13
  - https://github.com/renaissance-benchmarks/dependency-scala-smtlib/tree/renaissance/export

#### scala-stdlib/scala-kmeans (Scala 2.13)

- Updated Scala version (2.12.8 -> 2.13.5)

#### scala-stm/* (Scala 2.12)

- Updated Scala version (2.12.3 -> 2.12.13)
- The bundled `org.scala-stm.scala-stm-library` dependency will need updating for Scala 2.13

#### twitter-finagle/* (Scala 2.12)

- Updated Scala version (2.11.8 -> 2.12.13)
- Removed direct `com.twitter.twitter-server` dependency (19.4.0)
- Removed direct `com.twitter.common.metrics` dependency (0.0.39)
- Removed direct `com.twitter.common.io` dependency (0.0.69)
- Removed direct `com.twitter.util-events` dependency (7.0.0)
- Added dependency `com.google.guava.guava` dependency (19.0)
- Added dependency `commons-io.commons-io` dependency (2.4)

### Externally visible core/harness changes

- The `BenchmarkContext` has been changed to reduce clutter due to different methods for accessing benchmark parameters. The interface now provides only a single `parameter()` method returning an instance of `BenchmarkParameter`, which then provides the convenience methods parameter value conversion, allowing the two interfaces to evolve independently.
- The handling of scratch directories has been cleaned up, triggering a number of changes:
  - The `BenchmarkContext` interface provides the `scratchDirectory()` method, which provides a benchmark-specific scratch directory, which is removed when JVM exits. The other methods related to creating temporary directories will be removed when all benchmarks are updated to use the new interface.
  - The `DirUtils` class provides methods to recursively clean or delete directories so that benchmarks don't need to depend on the Apache `commons-io` library just for that.
  - The suite accepts two new command-line options `--scratch-base` and `--keep-scratch`. The first allows setting a base directory for scratch directories created by the suite, and the second prevents removal of those directories on JVM exit (for testing/debugging purposes).
  - The JMH wrapper (`JmhRenaissanceBenchmark`) provides similar functionality through system settings `org.renaissance.jmh.scratchBase` (path) and `org.renaissance.jmh.keepScratch` (boolean).
- The suite now checks the JVM version requirements selected for execution and excludes benchmarks that are incompatible with the current JVM from execution.
  - Consequently, the `--raw-list` option, which is intended for use in shell scripts, will only list benchmarks compatible with the JVM.
  - In both cases, the filtering can be disabled by using the `--no-jvm-check` option.
- The JMH wrapper can substitute (and execute) the `dummy-empty` benchmark in place of an incompatible benchmark (after issuing a textual warning). This is only useful for CI automation, because JMH does not provide a way to report/skip an incompatible benchmark without failure, and (unlike the normal Renaissance bundle) the JMH-enabled bundle cannot provide a list of compatible benchmarks. This behavior must be explicitly enabled by setting the `org.renaissance.jmh.fakeIncompatible` system property to `true`.

### Internal core/harness changes

- The `Launcher`, `RenaissanceSuite`, and the `JmhRenaissanceBenchmark` classes now create independent scratch directories to reflect the different contexts in which they execute.
- The `ModuleLoader` class is no longer a class-based (static) singleton, but is supposed to be instantiated and needs to be provided with a scratch directory into which it can write module libraries.
- The `BenchmarkInfo` class now provides a method to load a particular benchmark using a given `ModuleLoader` instance. The method has been moved from the `BenchmarkRegistry` class.


benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
```diff
@@ -36,52 +39,37 @@ final class Dotty extends Benchmark {
 
   private val outputPath = dottyPath.resolve("output")
 
-  private val sources: mutable.Buffer[String] = mutable.Buffer[String]()
+  private var dottyBaseArgs: Seq[String] = _
 
-  private var sourcePaths: mutable.Buffer[String] = null
+  private var dottyInvocations: Seq[Array[String]] = _
+
+  private val batchCompile = false
```
farquet, line 46: Sounds interesting to have this option exposed since you implemented both.

Could we turn this into a `@Parameter(name = "batch_compile", defaultValue =false)` ?

benchmarks/scala-sat/cafesat/build.sbt
```diff
@@ -1,41 +1,12 @@
...
-  out.setExecutable(true)
-  out
-}
-
-lazy val root = (project in file(""))
+lazy val scalaCafeSAT = (project in file("."))
   .settings(
     name := "CafeSat",
+    organization := "com.regblanc",
     version := "0.01",
```
farquet, line 7: Should we bump this value to avoid conflicts with a locally cached version of `0.01`?

benchmarks/scala-sat/src/main/scala/org.renaissance.scala.sat/ScalaDoku.scala
```diff
@@ -14,21 +14,24 @@ import org.renaissance.License
 
 object Solver {
 
-  /** Solver take a sudoku problem as a matrix of optional int,
+  /**
+   * Solver take a sudoku problem as a matrix of optional int,
```
farquet, line 18:
```
  /**
   * Solver takes a sudoku problem as a matrix of optional 
   * integers. It fills out the missing values of the matrix
   * to form a correct sudoku grid.
   */
```

farquet (commented):
This is an incredible work on both the quality and quantity sides!

I have a few minor comments based on the code review. I will now test the branch (and its new options) locally before giving the approval.

benchmarks/scala-dotty/build.sbt
```diff
@@ -5,11 +5,11 @@ lazy val scalaDotty = (project in file("."))
...
     organization := (organization in renaissanceCore).value,
-    scalafmtConfig := Some(file(".scalafmt.conf")),
-    scalaVersion := "2.12.8",
+    scalaVersion := "2.13.5",
+    scalacOptions += "-Ytasty-reader",
     libraryDependencies ++= Seq(
-      "commons-io" % "commons-io" % "2.6",
-      "ch.epfl.lamp" % "dotty-compiler_0.12" % "0.12.0"
+      "org.scala-lang.modules" %% "scala-collection-compat" % "2.4.2",
+      "org.scala-lang" % "scala3-compiler_3.0.0-RC1" % "3.0.0-RC1"
```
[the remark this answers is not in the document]
lbulej: I think I actually tried RC2, but the compilation fails, because (I think) the Tasty reader in Scala 2.13.5 is too old to read the jars produced by RC2. Actually, there is also RC3 on Maven central and it has the same problem, which is why I settled for RC1 back then. I think that we could wait for 3.0.0 (final) and hopefully, the updated Tasty reader will find its way into Scala 2.13.6.

Here's the error message I get with RC2 (and RC3):
```
[error] error while loading package, class file '/home/pallas/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3.0.0-RC2/3.0.0-RC2/scala3-compiler_3.0.0-RC2-3.0.0-RC2.jar(dotty/tools/package.class)' is broken
[error] (class scala.tools.tasty.UnpickleException/TASTy signature has wrong version.
[error]  expected: {majorVersion: 28, minorVersion: 0 [unstable release: 1]}
[error]  found   : {majorVersion: 28, minorVersion: 0 [unstable release: 3]}
[error] 
[error] This TASTy file was produced by an unstable release.
[error] To read this TASTy file, your tooling must be at the same version.
[error] The TASTy file was produced by Scala 3.0.0-RC2-nonbootstrapped.
[error] Note that your tooling is currently using an unstable TASTy version.)
[error] error while loading Main, class file '/home/pallas/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3.0.0-RC2/3.0.0-RC2/scala3-compiler_3.0.0-RC2-3.0.0-RC2.jar(dotty/tools/dotc/Main.class)' is broken
[error] (class scala.tools.tasty.UnpickleException/TASTy signature has wrong version.
[error]  expected: {majorVersion: 28, minorVersion: 0 [unstable release: 1]}
[error]  found   : {majorVersion: 28, minorVersion: 0 [unstable release: 3]}
[error] 
[error] This TASTy file was produced by an unstable release.
[error] To read this TASTy file, your tooling must be at the same version.
[error] The TASTy file was produced by Scala 3.0.0-RC2-nonbootstrapped.
[error] Note that your tooling is currently using an unstable TASTy version.)
```

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
lbulej: Can do, but it cannot be enabled for the current set of sources -- they have name clashes and are meant to be compiled individually. This is meant to be enabled for the `scalap` sources (that are not yet in). If I remember correctly, I was getting compilation warnings and wanted to investigate them.

benchmarks/scala-sat/cafesat/build.sbt
lbulej: Working on it. Need to add support for it into `dependency-cafesat` so that Renaissance contains a snapshot of that repository.

benchmarks/scala-sat/src/main/scala/org.renaissance.scala.sat/ScalaDoku.scala
lbulej: Done.

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
lbulej: Done. It did trigger a change in how benchmark access parameters, so that I can avoid changing the `BenchmarkContext` interface when I want to add new parameter type :-)

benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala
```diff
@@ -22,14 +22,26 @@ trait SparkUtil {
...
+    // However, setting spark.driver.bindAddress to "127.0.0.1" does not
+    // seem to work on Spark 2.4.7, whereas setting spark.driver.host to
+    // "localhost" or "127.0.0.1" does, so does setting the SPARK_LOCAL_IP
+    // environment variable (but we cannot do it from here).
+    //
     val conf = new SparkConf()
       .setAppName(benchName)
       .setMaster(s"local[$threadsPerExecutor]")
+      .set("spark.driver.host", "localhost")
+      .set("spark.driver.bindAddress", "127.0.0.1")
```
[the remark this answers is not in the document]
lbulej: Yeah, I despise software that tries to second-guess my network setup :-)

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
```diff
@@ -24,14 +24,19 @@ import scala.collection.compat._
 @Summary("Runs the Dotty compiler on a set of source code files.")
 @Licenses(Array(License.BSD3))
 @Repetitions(50)
+@Parameter(
+  name = "batch_compilation",
```
axel22, line 28: Minor: might want to consider dashes in parameter names, since it's easier to write.

axel22 (approved):

benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
lbulej: They are not meant to be specified on the command line -- they are mostly written down once for every configuration, so it's relatively infrequent. And we would have to redo all the parameter names :-)

benchmarks/scala-sat/cafesat/build.sbt
lbulej: That said, I'm not sure how it could conflict with a locally cached version, when we don't have it as a (maven-style) dependency, we don't publish it locally (yet), and it does not have an artifact on Maven central (the `cafesat` repo does not contain any version tag, so the version `0.01` just represents something experimental). But I think we should eventually switch to publishing dependencies locally, so I'll add some way of identifying the specific version w.r.t. our snapshot, similar to what we have in `dependency-scala-smtlib`.

lbulej: Done.

farquet (approved):
After careful local experimenting with all features from the branch, everything works like a charm as expected.

Feel free to merge it such that we can iterate on smaller PRs for upcoming changes :)

lbulej:
Thanks for taking the time to review and experiment with this pile of changes! I'll try to go back to much smaller PRs :-)

PR #246, opened 2021-04-30
Remove old temporary directory API from benchmark context

lbulej:
Updates the `apache-spark` and `dbshootout` benchmarks to use the `BenchmarkContext.scratchDirectory()` method, so that we can get rid of the old methods.

The other change moves the resolution of the benchmark scratch directory into a single place so that we don't have to duplicate policy-level decisions in the harness and the JMH wrapper.

There is another set of changes (#248) that makes the `apache-spark` benchmarks actually **use** the scratch directory and stop preparing their inputs (and dumping their outputs) into the `target` subdirectory. Similar change will be done for `scala-dotty`.


farquet (approved):
LGTM!

PR #247, opened 2021-04-30
Update Spark to version 3.1.1

lbulej:
This skips version 3.0.2 and goes directly to 3.1.1. There are no compilation issues and the apache-spark benchmarks execute normally. It might (but I did not check) with JDK compatibility issues. The attached CSV files show the times of the first iteration on OpenJDK 1.8.0_282 (just to show that it does not crash).

[spark301.csv](https://github.com/renaissance-benchmarks/renaissance/files/6404991/spark301.csv)
[spark311.csv](https://github.com/renaissance-benchmarks/renaissance/files/6404992/spark311.csv)

I think that until we freeze for a release, we should stay up-to-date whenever possible.


farquet (approved):
Cool, that's very up to date :)

lbulej:
Do you have any idea what's wrong with the OpenJDK9 on Mac? For some reason, the CI cannot move past the bundle stage on that platform due to some certificate issues when downloading something as rudimentary as sbt-git-javadoc. I have restarted the builds several times now, but to no avail.

lbulej:
A bit of investigation (but no final solution yet) in #249

farquet (approved):
LGTM

PR #250, opened 2021-04-30
Fix issues with outdated JDK trust store on OS X

### LINKED ISSUE #249: Outdated Oracle JDK9 trust store on Mac
Recent PRs have trouble passing the Travis bundle stage with Oracle JDK9 on MacOS X.

The problem looks like an outdated trust store:
- https://ao.gl/solved-pkix-path-building-failed-sun-security-provider-certpath-suncertpathbuilderexception-unable-to-find-valid-certification-path-to-requested-target/
- http://littlebigextra.com/how-to-fix-pkix-path-building-failed-sun-security-provider-certpath-suncertpathbuilderexception/
- https://confluence.atlassian.com/kb/unable-to-connect-to-ssl-services-due-to-pkix-path-building-failed-error-779355358.html

We can try to include the missing certificates in the repository and install them during Travis build:
```shell
find ./certs -name '*.pem' | while read CERT; do
    keytool -importcert -trustcacerts -keystore "$JAVA_HOME/jre/lib/security/cacerts" -storepass changeit -alias $(basename "$CERT" .pem) -file "$CERT" -noprompt
done
```

However, I'm not sure if we have the permissions to modify that file.

Alternatively, we would need to copy the `cacerts` file from the JDK system directory, modify it, and then use the `-Djavax.net.ssl.trustStore` property to point to the modified `cacerts` file. This might be amenable to integration in the `sbt` launch script and startup configuration (or creating a .sbt file with the JVM option), because `sbt` is the affected party here.

lbulej:
Fixes #249 caused by an outdated JDK trust store on OS X when using `xcode9.3` as `osx_image`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ jobs:
 
     - <<: *bundle
       os: osx
-      osx_image: xcode9.3
+      osx_image: xcode9.4
       jdk: oraclejdk9
     - <<: *bundle
       jdk: openjdk8
@@ -83,7 +83,7 @@ jobs:
 
     - <<: *bench
       os: osx
-      osx_image: xcode9.3
+      osx_image: xcode9.4
       jdk: oraclejdk9
     - <<: *bench
       jdk: openjdk8


lbulej:
Cherrypicking the change on top of #247 does not help. We might actually need to play with the trust store directly :-(

lbulej:
Superseded by #253

PR #252, opened 2021-05-03
Fix wrong dotty benchmark parameter and update dotty to use scratch directory

### LINKED ISSUE #251: Dotty benchmark is failing on master
As one can see in the Travis job (and that I can reproduce locally), the failure is:

```
====== dotty (scala-dotty) [test], iteration 0 started ======
GC before operation: completed in 38.139 ms, heap usage 62.500 MB -> 9.634 MB.
-- [E161] Naming Error: target/dotty/src/implicit-scope-loop.scala:4:6 ---------
4 |trait A[T] extends B
  |^^^^^^^^^^^^^^^^^^^^
  |A is already defined as class A in target/dotty/src/compound.scala
-- [E161] Naming Error: target/dotty/src/implicit-scope-loop.scala:5:6 ---------
5 |trait B extends Dummy[A[Int]]
  |^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |B is already defined as class B in target/dotty/src/compound.scala
-- [E161] Naming Error: target/dotty/src/implicit-scope-loop.scala:11:0 --------
11 |object Test {
   |^
   |Test is already defined as object Test in target/dotty/src/conformsWild.scala
-- [E161] Naming Error: target/dotty/src/implicitNums.scala:1:0 ----------------
1 |object Test {
  |^
  |Test is already defined as object Test in target/dotty/src/conformsWild.scala
-- [E161] Naming Error: target/dotty/src/inf.scala:1:0 -------------------------
1 |object Test {
  |^
  |Test is already defined as object Test in target/dotty/src/conformsWild.scala
-- [E161] Naming Error: target/dotty/src/interfaceObject.scala:1:6 -------------
1 |trait A{
  |^
  |A is already defined as class A in target/dotty/src/compound.scala
2 |  object O
3 |}
-- [E161] Naming Error: target/dotty/src/interfaceObject.scala:4:6 -------------
4 |class B extends A {
  |^
  |B is already defined as class B in target/dotty/src/compound.scala
5 |  def foo = O
6 |}
-- [E161] Naming Error: target/dotty/src/tcpoly_overloaded.scala:12:6 ----------
12 |trait Test {
   |^
   |Test is already defined as class Test in target/dotty/src/compound.scala
13 |  def moo: MList[Int]
14 |  class MList[T](el: T) extends Monad[T, List, [X] =>> Any] {
15 |    def flatMap[S <: RBound[S], RContainer[x <: RBound[x]], RBound[_],
16 |            Result[x <: RBound[x]] <: Monad[x, RContainer, RBound]]
17 |            (f: T => Result[S]): Result[S] = sys.error("foo")
18 |    def flatMap[S <: RBound[S], RContainer[x <: RBound[x]], RBound[_],
19 |            Result[x <: RBound[x]] <: Monad[x, RContainer, RBound]]
20 |            (f: T => Result[S], foo: String): Result[S]  = sys.error("foo")
21 |    def flatMap[S]
22 |            (f: T => List[S], foo: Int): List[S]  = sys.error("foo")
23 |  }
24 |  val l: MList[String] = moo.flatMap[String, List, [X] =>> Any, MList]((x: Int) => new MList("String"))
25 |}
-- [E161] Naming Error: target/dotty/src/tcpoly_variance_pos.scala:1:6 ---------
1 |class A[m[+x]] {
  |^
  |A is already defined as class A in target/dotty/src/compound.scala
2 | def str: m[Object] = sys.error("foo")
3 |}
-- [E161] Naming Error: target/dotty/src/tcpoly_variance_pos.scala:5:6 ---------
5 |class B[m[+x]] extends A[m] {
  |^
  |B is already defined as class B in target/dotty/src/compound.scala
6 | override def str: m[String]  = sys.error("foo")
7 |}
-- [E161] Naming Error: target/dotty/src/tparam_inf.scala:14:0 -----------------
14 |object Test {
   |^
   |Test is already defined as object Test in target/dotty/src/conformsWild.scala
-- [E161] Naming Error: target/dotty/src/wildcardBoundInference.scala:35:6 -----
35 |class Test {
   |^
   |Test is already defined as class Test in target/dotty/src/compound.scala
36 |  class MyExist extends ExistF[MyExist]
37 |  // SI-8197, SI-6169: java infers the bounds of existentials, so we have to as well now that SI-1786 is fixed...
38 |  def stringy: Exist[_ <: String] = (new Exist[String]).foo
39 |  // def fbounded: (ExistF[t] forSome {type t <: ExistF[t] }) = (new MyExist).foo
40 |  def indir: ExistIndir[_ <: String, _ <: String] = (new ExistIndir[String, String]).foo
41 |}
-- [E161] Naming Error: target/dotty/src/conformsWild.scala:1:0 ----------------
1 |object Test {
  |^
  |Test clashes with class Test in target/dotty/src/compound.scala; the two must be defined together
-- [E161] Naming Error: target/dotty/src/implicit-scope-loop.scala:6:0 ---------
6 |object B {
  |^
  |B clashes with class B in target/dotty/src/compound.scala; the two must be defined together
14 errors found
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== dotty (scala-dotty) [test], iteration 0 completed (4279.592 ms) ======
```
https://travis-ci.org/github/renaissance-benchmarks/renaissance/jobs/768860765

As mentioned in #242, there is the plan to replace the scala sources to compile. So hopefully the error goes away just by doing this.

Looks like result validation could have helped catch that though :D
```
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
```

lbulej:
This happens to do two things.

- First of all, it fixes #251 which was a result of introducing the `batch_compilation` benchmark parameter with a wrong default. The batch compilation was never meant for the testing sources we currently use because they use conflicting names and the files are meant to be compiled one by one.
- Second, it makes the benchmark use the benchmar-specific scratch directory instead of using the hard-coded path `target/dotty` in the current directory.

There are other things that need to be done with `dotty`, probably in other PR:

- Switch to the `scalap` sources. The compilation actually needs the `org.scala-lang:scala-compiler:2.13.0` dependency. There are also some warnings during compilation and we need to decide how to handle them.
- More importantly, we should run the compiler `Driver` class with at least a simple reporter, so that we get information about compilation errors.
- Finally, we need to add validation, possibly by hashing the produced class files.


farquet (approved):
LGTM

Issue #243, opened 2021-04-26
Try GitHub actions


vhotspur:
Seems to be more flexible than Travis.

Opening mainly to track progress.

PR #253, opened 2021-05-03
Switch to git-based versioning and rework the CI to follow suit

### LINKED ISSUE #249: Outdated Oracle JDK9 trust store on Mac

lbulej:
Switch the build system to use git-based versioning, where we use `git describe` to determine Renaissance version. The current version number is derived from the latest (annotated) tag and its distance from the branch HEAD. I added those tags for the past releases. A release built from the tagged commit will have a "clean" version number, others will include the hash of the build commit. This has two immediate effects:
- We don't have to change the version number in multiple places (`renaissance-core/version.sbt`, `.travis.yml`, and `README.md`) and development builds have somewhat traceable version number.
- The pre-push and CI scripts can determine the current version by stripping the initial `v` from the output of `git describe`.

This, in turn, made it possible to take the contents of the `script` section in `.travis.yml` and turn it into a set of CI actions in separate scripts that are just listed in `.travis.yml`. I actually planned to do something like that for some time already, but could not get around to it, until issue #249 came around. A few changes happened along the way:
- We now do the pre-flight checks on OS X (so far the most fragile platform) so that we don't waste time building Renaissance when we can't even download the dependencies.
- We don't test with Oracle JDK9 on Linux, because we already do it on OS X. Instead, we test JDK8 from Oracle and OpenJDK, and JDK11 from Oracle, OpenJDK, and OpenJ9 (again).
- We use SBT in client mode, which leaves a server running and improves SBT startup on subsequent invocations (we have 2 or 3 of those).

This also fixes #249, so that Oracle JDK9 builds on OS X stop failing due to the inability to download dependencies (which has become really annoying because it was blocking all the PRs while wasting time building Renaissance on other platforms). We now use our own `cacerts` key store which includes the certificate chain that is missing on OS X, but we can stop doing that when it's not needed anymore. 

Hopefully, the scripts can be also used with GitHub Actions (#243).

With respect to versioning, the only thing that *should* be updated on release is the `git.baseVersion` setting in `renaissance-core/version.sbt`, which provides a base (upcoming) version number for git-less situations, i.e., when building the sources outside the git tree.  This also necessitated an update of the `reactors` dependency to support git-less mode of operation.

The final piece of the puzzle is an updated pre-push hook, which also checks the generated markdown files so that forgetting to update them does not block a PR after push when it's something that can be found out locally. The only downside is that it requires building the Renaissance bundle before push, but that should not be a problem. We don't use the client mode in the pre-push hook, but perhaps we should.


vhotspur (approved):
LGTM, thanks for the CI scripts!

PR #257, opened 2021-05-12
Use test configuration in CI runs of the JMH bundle (and add convenience scripts)

lbulej:
This modifies the JMH wrapper to be able to use other configurations, changes the CI script that runs the JMH bundle to use the `test` configuration, and ensures that we pass the right JVM arguments to the JMH-forked VM.  These are relatively small changes on top of `master`, not a long-running branch (what #255 should have been but wasn't). And yes, it also tacks on two convenience scripts to run the current build of Renaissance and the JMH bundle, which is useful during development.


vhotspur (approved):
LGTM.

PR #259, opened 2021-05-30
Update SBT to 1.5.2

lbulej:
Removes usage of deprecated scoping syntax and updates a few plugins so that we are not left behind. It should also allow us to use "3.0.0" as `scalaVersion` when the need comes.


farquet (approved):
LGTM

PR #261, opened 2021-06-07
finagle-chirper: read with UTF-8 explicitly

vhotspur:
Otherwise, reading of the `new-years-resolution.csv` will fail on systems where UTF-8 is not the default charset.


farquet (approved):
LGTM

PR #262, opened 2021-06-07
Allow overriding benchmark parameters

lbulej:
Adds `-o` or `--override` option accepting a `<name>=<value>` pair. The override is applied to the selected configuration and benchmarks. An alternative would be to allow `<name>`  to be a full name (or a regexp matching the name) of a property found in the `benchmarks.properties`  file.

At runtime, if the selected benchmark configuration has an overridden parameter, the suite prints the parameter together with the old and new value before running the benchmark and shows the displayed configuration name with a `modified-` prefix.

Interestingly, finishing the implementation after the initial proof-of-concept required a bit of refactoring which also enabled consolidation of some of the services provided by the suite core classes into the `BenchmarkSuite` class, avoiding duplication of some of the functionality in the `RenaissanceSuite` and `JmhRenaissanceBenchmark`.  More details can be found in the commit messages.


lbulej:
Interesting, how such an innocent feature can amplify the lack of proper abstractions... Who's up for round 2?

lbulej:
BTW, travis-ci.org is now officially gone and the first reactions to travis-ci.com (from free users) seem to be... underwhelming...

renaissance-harness/src/main/java/org/renaissance/harness/ExecutionDriver.java
```diff
@@ -1,103 +1,112 @@
...
-    this.scratchRoot = scratchRoot;
+    this.benchmarkContext = context;
+    this.benchmark = benchmark;
+    this.eventDispatcher = dispatcher;
+    this.executionPolicy = policy;
     this.vmStartNanos = vmStartNanos;
+
+    this.benchmarkName = context.benchmarkName();
+    this.beforeEachMessageFormat = String.format(
+      "====== %s (%s) [%s], iteration %%d started ======\n",
```
farquet, line 61: Oh, I got confused by the escaped percent, but you're actually just preparing a string to format with a string format :)

farquet (approved):
That's indeed a much nicer design with better abstractions for suites and benchmarks to run.
Nice :)

renaissance-harness/src/main/java/org/renaissance/harness/ExecutionDriver.java
lbulej: I wanted to preformat the stable things to avoid passing it down the print functions over and over.

PR #263, opened 2021-06-07
Update scala-dotty to use scalap sources and do hash-based validation

lbulej:
- Updates `scala-dotty` to use `scalap` sources. After some cleanups, they produce just 6 warnings related to implicits.
- Adds validation phase that requires compilation to complete without any errors (warnings are tolerated).
- Checksums generated .tasty files in the validation phase with the exception of two files that end up containing the current working directory path.
- Removes the `batch_compilation` parameter, because it does not make sense to compile the sources individually.
- Removes the `expected_tasty_hash` parameter, because there is no need to parameterize it (yet).


lbulej:
Hmm, it seems that even the .tasty files are not stable and depend on who-knows-what. Why would anyone want to do that without providing an option to stop such a madness?

farquet:
Maybe @smarter could give an input here. Is there a way to have the Tasty files always be consistently compiled the same way such that hashes of those files could be used to verify correct compilation?

lbulej:
Further investigation shows that it's the current working directory that makes it into some of the .tasty files (not all, only `scala/tools/scalap/Classfile.tasty` and `scala/tools/scalap/scalax/rules/ByteCode.tasty`).

The path that actually gets embedded in the .tasty files is in the form of `<cwd>/library/src/scala/runtime/stdlibPatches/Predef.scala` which never even existed during compilation, so it does not make sense to concatenate with the current working directory.

Setting the `user.dir` system property before calling Dotty has no effect, neither does using the `-sourcepath` option, nor does using an absolute path with the `-sourceroot` option.

One way to go forward is to just exclude these two files from the checksum.

smarter:
> Is there a way to have the Tasty files always be consistently compiled the same way such that hashes of those files could be used to verify correct compilation?

They're supposed to be always the same (and we have tests for that: https://github.com/lampepfl/dotty/tree/master/tests/idempotency), if they're not it's a bug

@lbulej could you open an issue at https://github.com/lampepfl/dotty/issues with reproduction steps for this?

lbulej:
> > Is there a way to have the Tasty files always be consistently compiled the same way such that hashes of those files could be used to verify correct compilation?
> 
> They're supposed to be always the same (and we have tests for that: https://github.com/lampepfl/dotty/tree/master/tests/idempotency), if they're not it's a bug
> 
> @lbulej could you open an issue at https://github.com/lampepfl/dotty/issues with reproduction steps for this?

Thanks for the response @smarter, I'll see if I can reproduce it outside Renaissance just with `scalac` so that I can provide more useful input.

PR #267, opened 2021-06-12
Make build.sbt (and friends) a bit easier to understand

lbulej:
This does several things in an attempt to streamline the build process:
- Marks dependency on `renaissance-core` in other modules as `provided`.  This is a dependency that is expected to be provided by the *runtime*, which is exactly what we do. As a result, the `renaissance-core` JAR file is excluded from the runtime dependencies of other modules and is not repeatedly copied into the final bundle.
- Dismantles the `projectJars` and `jarsAndListGenerator` tasks which together did too many things into separate tasks responsible for generating the `modules.properties` and `benchmarks.properties` files and for copying the runtime dependencies into the `resource_managed` directory of the root project. 
- Factors the inline definition of the task to generate JMH wrappers out of the `renaissanceJmh` project configuration and makes it a separate task that also relies on the `renaissanceBenchmarks` projects to select benchmarks for which to generate the wrappers, reducing `RenaissanceJmh` to just a single method that generates the wrapper.

### DIFF
--- a/benchmarks/actors-akka/build.sbt
+++ b/benchmarks/actors-akka/build.sbt
@@ -12,5 +12,5 @@ lazy val actorsAkka = (project in file("."))
     )
   )
   .dependsOn(
-    renaissanceCore
+    renaissanceCore % "provided"
   )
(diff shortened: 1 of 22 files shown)


farquet (approved):
Nice cleanup

PR #270, opened 2021-06-17
Make the benchmark primary groups more user-friendly

lbulej:
For now, we retain the original groups that mostly correspond to benchmark modules, so that we don't break scripts that might have been using them. We also display the primary group in benchmark output (instead of benchmark module), which I believe we always intended, but there was no need to differentiate between benchmark module and benchmark group back then.


farquet (approved):
Makes sense!

Issue #265, opened 2021-06-08
Spark benchmarks fail on Windows


vhotspur:
```
java.lang.Exception: Wrong winutils.exe size: expected 109568, written 112640
        at org.renaissance.apache.spark.SparkUtil.setUpHadoop(SparkUtil.scala:124)
        at org.renaissance.apache.spark.SparkUtil.setUpSparkContext(SparkUtil.scala:42)
        at org.renaissance.apache.spark.SparkUtil.setUpSparkContext$(SparkUtil.scala:38)
        at org.renaissance.apache.spark.Als.setUpSparkContext(Als.scala:50)
        at org.renaissance.apache.spark.Als.setUpBeforeAll(Als.scala:95)
        at org.renaissance.harness.ExecutionDriver.executeBenchmark(ExecutionDriver.java:59)
        at org.renaissance.harness.RenaissanceSuite$.$anonfun$runBenchmarks$1(RenaissanceSuite.scala:133)
        at org.renaissance.harness.RenaissanceSuite$.$anonfun$runBenchmarks$1$adapted(RenaissanceSuite.scala:127)
        at scala.collection.immutable.List.foreach(List.scala:333)
        at org.renaissance.harness.RenaissanceSuite$.runBenchmarks(RenaissanceSuite.scala:127)
        at org.renaissance.harness.RenaissanceSuite$.main(RenaissanceSuite.scala:104)
        at org.renaissance.harness.RenaissanceSuite.main(RenaissanceSuite.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.renaissance.core.Launcher.loadAndInvokeHarnessClass(Launcher.java:114)
        at org.renaissance.core.Launcher.launchHarnessClass(Launcher.java:73)
        at org.renaissance.core.Launcher.main(Launcher.java:37)
```

Probably due to update of `winutils.exe` resource.

lbulej:
You are right. I'll bump Spark and update everything.

vhotspur:
Please, see my `issues/265` branch.

lbulej:
[...]

PR #271, opened 2021-06-18
Enable GitHub Actions

vhotspur:
This PR enables use of GitHub Actions for CI. (Cleaned-up version of the `issue/243` branch. Travis CI will be removed as part of another PR.)

Current status:

* Checks (encoding, formatting) ... ok.
* Linux builds (and runs): JDK 8, 9, 10, 11, Openj9 on JDK 8 and 11 ... ok (ignoring spurious failures of the build servers).
* Linux builds (and runs): JDK 12+ .. failing.
* MacOS builds (and runs): JDK 8 and 11 ... ok.
* Windows builds (and runs): failing (see #265).

This is a base version, we can later switch to a more complex workflow.

GHA also allows for artifacts but the pricing is not clear to me at the moment (i.e., no artifacts are stored at this moment).

To decide before merge:

 * Remove the JDK12+ builds at the moment (they fail because of db-shootout)? I hope I will fix the Windows builds soon so these I would like to keep.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,10 +48,10 @@ jobs:
           - openjdk9
           - openjdk10
           - openjdk11
-          - openjdk12
+          #- openjdk12
           - openj9-openjdk8
           - openj9-openjdk11
-          - openj9-openjdk16
+          #- openj9-openjdk16
     runs-on: ubuntu-latest
     container: "renaissancebench/buildenv:${{ matrix.image }}"
     continue-on-error: true
@@ -100,7 +100,9 @@ jobs:
     strategy:
       matrix:
         java: [ '8', '11' ] #, '13', '15' ]
-        os: [ windows-latest, macos-latest ]
+        os:
+          #- windows-latest
+          - macos-latest
     runs-on: ${{ matrix.os }}
     continue-on-error: true
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,7 @@ jobs:
           path: |
             ~/.ivy2/cache
             ~/.sbt
+            ~/.cache/coursier
           key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
       - name: Check file encoding"
         shell: bash
@@ -66,6 +67,7 @@ jobs:
           path: |
             ~/.ivy2/cache
             ~/.sbt
+            ~/.cache/coursier
           key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
 
       - name: Environment configuration
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,147 @@
+name: build
+on: [push, pull_request]
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    container: "renaissancebench/buildenv:openjdk8"
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Cache SBT
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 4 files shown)


lbulej:
> 
> To decide before merge:
> 
>     * Remove the JDK12+ builds at the moment (they fail because of db-shootout)? I hope I will fix the Windows builds soon so these I would like to keep.

Or we could mark `db-shootout` as supporting JDK 11 at most. We limit `neo4j` to JDK 15 (I think) in the same way.

vhotspur:
> > To decide before merge:
> > ```
> > * Remove the JDK12+ builds at the moment (they fail because of db-shootout)? 
> > ```
> 
> Or we could mark `db-shootout` as supporting JDK 11 at most. We limit `neo4j` to JDK 15 (I think) in the same way.

I hope we will be able to upgrade it as the newer versions of the OpenHFT libraries support JDK12+ (though the dependencies specifications for `db-shoutout` are quite [magical to me](https://github.com/renaissance-benchmarks/renaissance/blob/81f18e0fbeb33fbaa56fed4a211f1738002b25ba/benchmarks/database/build.sbt#L17)).

lbulej:
I think this would be best merged in stages, prioritizing JDK 8 and JDK 11 (LTS) builds, which can go in immediately so we get back to having CI.

Then you can follow up with Windows support (not necessarily a primary target, but it can expose some potentially unhealthy assumptions we make).

Finally, we can add other JDKs if we can afford to run them, which will force us to fill out the JVM-supported column of the table in `README.md` :-)

vhotspur:
> I think this would be best merged in stages, prioritizing JDK 8 and JDK 11 (LTS) builds, which can go in immediately so we get back to having CI.

Okay, here it is: 0ad7218ee06a8f62d2eb740e961a468df0451e13 (and build results should be visible on this page).

lbulej:
> I hope we will be able to upgrade it as the newer versions of the OpenHFT libraries support JDK12+ (though the dependencies specifications for `db-shoutout` are quite [magical to me](https://github.com/renaissance-benchmarks/renaissance/blob/81f18e0fbeb33fbaa56fed4a211f1738002b25ba/benchmarks/database/build.sbt#L17)).

Hmm, hmm, it seems like were are trying to get very specific versions of individual components and the exclusions are meant to avoid overriding that.

This one is clear:
```
net.openhft" % "chronicle-core" % "2.17.2",
```

And these want newer `chronicle-bytes` and `chronicle-threads`, without pulling in newer `chronicle-core`:
```
"net.openhft" % "chronicle-bytes" % "2.17.7" exclude ("net.openhft", "chronicle-core"),
"net.openhft" % "chronicle-threads" % "2.17.1" exclude ("net.openhft", "chronicle-core"),
```

And the last wants a newer `chronicle-map` without overriding the specialized versions above.
But I don't know why we can't just pull a single version.

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+    runs-on: ubuntu-latest
+    container: "renaissancebench/buildenv:openjdk8"
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Cache SBT
+        uses: actions/cache@v2
+        with:
```
lbulej, line 14: If this is caching, we might want to also include `$HOME/.cache/coursier`

lbulej (approved):
LGTM.

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+          path: |
+            ~/.ivy2/cache
```
lbulej, line 16: If this is caching, we might want to also include `$HOME/.cache/coursier`

vhotspur: Added, thanks!

Issue #272, opened 2021-06-21
Windows builds broken


vhotspur:
Execution ends with the following trace:

```
[2021-06-21T02:59:16.727-0700] org.renaissance.core (org.renaissance.core.Launcher main)
SEVERE: Failed to load harness: org.renaissance.core.ModuleLoader$ModuleLoadingException: could not extract resource 'renaissance-harness\renaissance-harness_
2.13-0.11.0-252-g32aeaaf.jar' into 'C:\Users\vhotspur\renaissance-issue-265\launcher-025916-2120882362573410541\renaissance-harness\lib\renaissance-harness\re
naissance-harness_2.13-0.11.0-252-g32aeaaf.jar': resource not found
        at org.renaissance.core.ModuleLoader.extractResources(ModuleLoader.java:224)
        at org.renaissance.core.ModuleLoader.createClassLoaderForModule(ModuleLoader.java:159)
        at org.renaissance.core.Launcher.loadAndInvokeHarnessClass(Launcher.java:111)
        at org.renaissance.core.Launcher.launchHarnessClass(Launcher.java:73)
        at org.renaissance.core.Launcher.main(Launcher.java:37)
```

ceresek:
On a related note, does `--plugin` work for you ? I've had a problem with 510b3b9 (not master branch) because it tried to use a filename as a URL, which failed, and one cannot specify the protocol on the command line because the loader uses the separator. I've solved this with a quick hack to get things running but if it also impacts master we should take a look.

vhotspur:
> On a related note, does `--plugin` work for you ?

No (looked at  018c8e3). I will open a separate ticket.

PR #273, opened 2021-06-21
Attempted fix for #272 (broken Windows builds)

vhotspur:
The family of `getResource` functions requires [Unix-style paths](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getResource-java.lang.String-).

The filenames are sane here so this trivial fix should work but perhaps @lbulej could confirm that it will not break something else. Thanks!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,6 +9,7 @@ import java.nio.file.Path
 import java.nio.file.StandardCopyOption
 import java.util.Properties
 import scala.collection._
+import scala.jdk.CollectionConverters.asScalaIteratorConverter
 
 enablePlugins(GitBranchPrompt)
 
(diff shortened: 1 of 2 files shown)


lbulej:
This might break `ModuleLoader` but it depends on how `Path.resolve()` on Windows deals with a string that contains `/` as a separator. Let me see if I can extract the jars in a more robust way.

vhotspur:
> Let me see if I can extract the jars in a more robust way.

The extraction is not a problem: the issues is in creation of the `.properties` files where we should store the Unix-style path. Or am I missing something else here?

lbulej:
> > Let me see if I can extract the jars in a more robust way.
> 
> The extraction is not a problem: the issues is in creation of the `.properties` files where we should store the Unix-style path. Or am I missing something else here?

You are right. The extraction only used the last component of the name and resolved it against a directory path that was built properly from scratch, i.e., while respecting the platform separators.

lbulej (approved):
LGTM

vhotspur:
Thanks for the clean-up!

PR #274, opened 2021-06-21
Remove executor control and explicit repartitioning (and few other cleanups)

lbulej:
This was originally meant to only do one thing, but it ended up doing a little bit more as I stumbled upon other things.

## Changes related to Spark setup
- Removes the parameter to configure the number of Spark executors because it does not make sense in `local` mode
- Removes explicit partitioning (for now) from `als` and `movie-lens` benchmarks, leaving Spark to use the default which should correspond to the number of threads.
- Loads `page-rank` input using `SparkContext.textFile()` instead of calling `parallelize()` on a collection. This makes the benchmark use the same partitioning defaults as the other benchmarks.
- [x] Limits (caps) the number of threads in the Spark local executor instead of configuring the executor with a fixed number of threads. To reflect the change, the `spark_thread_count` parameter is replaced with `spark_thread_limit` parameter.

## Changes related to Spark machine learning API
- Avoid conversion from `RDD` to `DataFrame` on each benchmark operation in `log-regression`. Instead, we read the input data directly as a `DataFrame` which is what `LogisticRegression` consumes.
- Convert `naive-bayes` to use `ml.classification.NaiveBayes` instead of `mllib.classification.NaiveBayes`. This avoids internal conversion from `RDD` to `DataFrame` on each benchmark operation because the `mllib` version is just a wrapper around the `ml` version. Again, we read the input data directly as a `DataFrame` which is what `ml.classification.NaiveBayes` consumes.
- Drop usage of the deprecated `SQLContext` in `dec-tree`.

## Additional cleanups
- Removes duplicated parsing of movie ratings from `movie-lens`.
- Factors out duplicated input preparation from `dec-tree`, `log-regression`, and `naive-bayes`.
- Removes manual parsing of `libsvm` data from `log-regression` and `naive-bayes` and instead uses the reader provided by `SparkSession`.

### Do we want to play with input data partitioning?

Not now. Eventually, we want to scale workloads with the number of threads at which point we should analyze the benchmarks to find out reasonable input partitioning (or leave it on defaults).

### Should we try to avoid internal reshaping of input data?

  - For example, `mllib.ALS` uses `ml.ALS` and remaps `RDD[mllib.Rating]` to `RDD[ml.ALS.Rating]` before running the algorithm using the `train()` method from the `ml.ALS` companion object. Interestingly, even the `fit()` method from the `ml.ALS` class (the new API) remaps the input `Dataset` into an `RDD[ml.ALS.Rating]`.
  - The conversions are probably OK for common usage, which might use different instances of big data, and the conversion is only done once, but when benchmarking, repeating the conversion just makes things take longer time and consume significantly more memory.
  - For example, in the `naive-bayes` benchmark, which inflates the input 8000x, the repeated conversion did significantly inflate the running time and memory usage. This wasn't a problem in Renaissance 0.11 and is probably the result of upgrading to Spark 3, where some of the old API classes (MLlib) only provide a wrapper over new API classes (ML).
  - However, to avoid these conversions when we already have a properly typed `RDD[ml.ALS.Rating]`, we would have to use the `ml.ALS.train()` method directly, which is not something normal users would do.

For now, the consensus is that we should try to stick to the higher-level APIs that are more likely to be used by "normal" users.


ceresek:
Maybe also, do we want to support running benchmarks with non-local master (by having an option to use other master than `local`) ? (Not sure if that would even work.)

I did not measure the modified versions yet, but from the earlier data, almost no Spark benchmark benefits from using more than 6-8 threads (only naive-bayes does), and at least three benchmarks run into scalability issues with more than 6-8 threads (als, movie-lens, page-rank).

lbulej:
> Maybe also, do we want to support running benchmarks with non-local master (by having an option to use other master than `local`) ? (Not sure if that would even work.)

How do you imagine it to work? The benchmark could probably use a non-local master, but who would be responsible for starting the cluster nodes and the master node? We could try starting nodes locally on a reasonably big machine, but that would turn it into a multi-VM benchmark.

> 
> I did not measure the modified versions yet, but from the earlier data, almost no Spark benchmark benefits from using more than 6-8 threads (only naive-bayes does), and at least three benchmarks run into scalability issues with more than 6-8 threads (als, movie-lens, page-rank).

I've built a bundle that you might try running instead of the one that we run now (it's in the same directory). I have modified the run script to only iterate over thread counts and to produce filenames with `E0` to differentiate it from the other results. The plotting scripts will ignore those results because of that (for now).

ceresek:
> How do you imagine it to work? The benchmark could probably use a non-local master, but who would be responsible for starting the cluster nodes and the master node? We could try starting nodes locally on a reasonably big machine, but that would turn it into a multi-VM benchmark.

An entirely manual setup - if somebody wants to experiment, they could launch the cluster and tell Renaissance to use it. We would just provide the option to do so without compiling the sources.

> I've built a bundle that you might try running instead of the one that we run now (it's in the same directory). I have modified the run script to only iterate over thread counts and to produce filenames with `E0` to differentiate it from the other results. The plotting scripts will ignore those results because of that (for now).

Before I run this, could we please also fix the `--plugin` issue I've mentioned in my other comment with #272 ? I might be missing something, but it seems we're using `ModuleLoader.loadExtension` to load the plugin class, but that method expects a URL where the command line option provides a file name, and thus fails (and URL cannot be entered on the command line because `loadExtension` splits the input using `pathSeparator`, which is the same as protocol separator on Linux).

lbulej:
For testing, I added `als-ml` benchmark, which uses `ml.ALS` instead of `mllib.ALS`. Both do a conversion to RDD and then use 'ml.ALS` for the actual computation, but the MLlib version seems to be more efficient (it does RDD->RDD conversion instead of Dataset->RDD).

lbulej:
I have added few other changes but I think it's ready to be merged. We should set the thread limits in a separate PR, perhaps with a few accompanying plots. One thing that's left to decide is whether we keep `als` as is or we replace it with the `als-ml` variant. If we don't want to decide now, I can either remove it (or mask it from the harness in some way).

farquet:
> One thing that's left to decide is whether we keep als as is or we replace it with the als-ml variant. If we don't want to decide now, I can either remove it (or mask it from the harness in some way).

I personally like the fact that we keep benchmarks with both APIs, especially since the RDD approach is not deprecated and potentially used by people around the world. Moreover, it covers a different aspect of the framework (data movement I think). So my personal take is to stick to `als` for now since the ML variant is used in other benchmarks already. But I have no strong opinion on this.

Another option is to keep both. We have `mnemonics` and `par-mnemonics` which are variants of the same workload. Same for `scrabble` with its `rx-scrabble` counterpart.

So if I were to choose now, I would either keep `als` and hide `als-ml` or just have both around.

lbulej:
> 
> I personally like the fact that we keep benchmarks with both APIs, especially since the RDD approach is not deprecated and potentially used by people around the world. Moreover, it covers a different aspect of the framework (data movement I think). So my personal take is to stick to `als` for now since the ML variant is used in other benchmarks already. But I have no strong opinion on this.
> 
> Another option is to keep both. We have `mnemonics` and `par-mnemonics` which are variants of the same workload. Same for `scrabble` with its `rx-scrabble` counterpart.
> 
> So if I were to choose now, I would either keep `als` and hide `als-ml` or just have both around.

The situation with `als` and `als-ml` is kind of strange, because they both end up calling the same implementation from the `ml` (newer) package which, however, is still RDD-based. I guess that one day there will be a DataFrame-native implementation, but for the time being, both versions do a conversion to RDD (even the RDD-based wrapper in the `mllib` package). But if/when a DataFrame-based implementation arrives in `ml`, it will be the `mllib` version doing RDD->DataFrame conversion (which is why I ported `naive-bayes` to the `ml` classes which use DataFrame directly).

Anyway, the reason why I would prefer to pick only one is to avoid increasing the number of benchmarks without adding something really new, because one problem with `als` is that it uses the same algorithm as `movie-lens`, so it's kind of subsumed by it.

The difference is that `als` generates a huge random input data set and runs the algorithm once on it (without tweaking algorithm parameters), while `movie-lens` uses a smaller "real-world" data set and runs the algorithm on it multiple times, tweaking the algorithm parameters to find the best fit.

The fact that `als` generates its input data set may be useful later, especially if we decide to try scaling the workload size with the number of threads. But at this point, having two benchmarks that end up using the same implementation with the same data set does not seem too appealing. Interestingly, they do behave differently with a different number of threads, but I suspect that the difference is mostly due to the input data set conversions (DataFrame->RDD vs. RDD->RDD).

farquet:
Ok, makes sense.
So would we be well covered with different reasonable approaches by keeping `naive-bayes` (new), `als` (with wrapper) and `movie-lens` (different input data set approach)? What would be your suggestion otherwise?

ceresek (approved):

axel22:
I think that it would be good to keep both the new API and the old RDD API, since it still represents the workload that some people have, and even though it's being "phased out", it represents a pattern that may occur in libraries and programs, which is very useful for a VM implementer or a compiler developer to have.

lbulej:
> So would we be well covered with different reasonable approaches by keeping `naive-bayes` (new), `als` (with wrapper) and `movie-lens` (different input data set approach)? What would be your suggestion otherwise?

I would tend to use the `ml` version of `als`, which would get us to the following state:
- `als` would use the DF-based ML API from `ml`, but it would still use RDD internally (until the implementation in Spark is updated)
- `dec-tree` uses DF-based ML API from `ml`
- `log-regression` uses DF-based ML API from `ml`
- `naive-bayes` uses DF-based ML API from `ml`, because `mllib` version only wraps the `ml` version anyway
- `chi-square` uses RDD-based ML API from `mllib`
- `movie-lens` uses RDD-based ML API from `mllib`, ends up using the same ALS implementation in `ml`, but does different conversions
- `gauss-mix` uses RDD-based ML API from `mllib`
- `page-rank` uses plain RDD API

So it's kind of balanced when it comes to `ml` vs. `mllib`, but with a slight edge towards RDD :-)

lbulej:
> I think that it would be good to keep both the benchmarks with new API and with the old RDD API, since the old version still represents the workload that some people have, and even though it's being "phased out", it represents a pattern that may occur in libraries and programs, which is very useful for a VM implementer or a compiler developer to have.

That's right. However, my point is that when the `mllib` implementation is just a compatibility wrapper around the `ml` implementation, I think it's time to move on, because even though the workloads written against `mllib` will keep working, it will be the `ml` code that ends up executing, modulo some conversions. To actually exercise the original `mllib` code paths, we would have to use older Spark (and older Renaissance).

lbulej:
OK, so I guess I'll merge this now, so that I can open a new PR where we set thread limits based on the data from Petr and we can decide about `als` as the very last thing.

PR #275, opened 2021-06-21
dotty validation: force C-locale ordering

vhotspur:
This fixes the validation on Windows where the sorting was case insensitive.


lbulej:
Interesting... When I wrote it, I checked that `File` is actually `Comparable` and the `compareTo()` method should provide lexicographic ordering, but I probably finished reading too early to learn that it calls for a file-system-specific ordering which may ignore character case. Would that explain the failures?

lbulej (approved):
LGTM

PR #277, opened 2021-06-22
GitHub Actions: run on Windows too

vhotspur:
It is not perfect as the helper shell scripts are not used -- as on Linux and MacOS -- but they fail without any hint why (it seems to me that it is related to the SBT server/client feature but that is only a guess).

It would be nice to have it more Windows-only -- i.e. use `.bat` instead of Bash scripts --  but that can be changed later. I would like to have it in `master` ASAP to catch issues on Windows earlier. At the moment, it compiles and runs on Windows without any issues (up to JDK 11).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,14 +98,11 @@ jobs:
         shell: bash
         run: "tools/ci/bench-jmh.sh"
 
-  other:
+  macos:
     strategy:
       matrix:
         java: [ '8', '11' ] #, '13', '15' ]
-        os:
-          #- windows-latest
-          - macos-latest
-    runs-on: ${{ matrix.os }}
+    runs-on: macos-latest
     continue-on-error: true
     steps:
       - name: Git checkout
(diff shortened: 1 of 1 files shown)


lbulej (approved):
I'd say let's go for it. Regarding the .bat files -- would that be the recommended approach? I'd say if we can use bash, let's stick with bash, but we might try to factor out the calls to `ci_sbt --client` into `ci_cbt_client()` that will behave differently on Windows (i.e., not use the --client parameter, even though the bundled SBT does have a Windows version of the native client).

vhotspur:
> Regarding the .bat files -- would that be the recommended approach? I'd say if we can use bash, let's stick with bash

Honestly, I do not know. The point is that using `.bat` would not require any extra dependencies to be installed. I have no idea if every developer on Windows has Bash installed.

lbulej:
> > Regarding the .bat files -- would that be the recommended approach? I'd say if we can use bash, let's stick with bash
> 
> Honestly, I do not know. The point is that using `.bat` would not require any extra dependencies to be installed. I have no idea if every developer on Windows has Bash installed.

Let's focus on being able to *build* and *run* on Windows (current state I believe) and worry about *developing* on Windows when we find a Windows-based developer who wants to contribute to Renaissance.

PR #280, opened 2021-06-25
CI: test with plugins too

vhotspur:
Ensure we build our plugins regularly too.


lbulej (approved):
There is no reason to postpone this now. LGTM.

PR #281, opened 2021-07-01
Temporarily limit supported JVM in 'dbshootout' to version 11

lbulej:
This is to avoid out-of-the-box crashes on JVM 12 before we can resolve issue #163.


farquet (approved):

vhotspur:
That was quick :-). I will prepare a JDK12 setup for GHA.

lbulej:
> That was quick :-). I will prepare a JDK12 setup for GHA.

Sorry,  I'm eager to close loose ends for now so that we can move to a new (post-release) era ;-)

Issue #260, opened 2021-06-07
Spark worker scaling parameters


ceresek:
Currently, the performance of the Spark benchmarks does not change with the configured number of executors, except for ALS, which partitions the input data based on the configuration. This may be a relevant note in the Spark documentation:

> Spark properties mainly can be divided into two kinds: one is related to deploy, like “spark.driver.memory”, “spark.executor.instances”, this kind of properties may not be affected when setting programmatically through SparkConf in runtime, or the behavior is depending on which cluster manager and deploy mode you choose, so it would be suggested to set through configuration file or spark-submit command line options; another is mainly related to Spark runtime control, like “spark.task.maxFailures”, this kind of properties can be set in either way.

This is quite vague, but may explain why our code (https://github.com/renaissance-benchmarks/renaissance/blob/510b3b9e8f01dc397f35b64a7b5f8a943b75d012/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala#L66) does not behave as expected.

ceresek:
![workers-page-rank](https://user-images.githubusercontent.com/6158703/121003095-6abcb400-c78d-11eb-8fa3-569bf2a880d8.png)

ceresek:
... seeing how our Spark instance is hardcoded to run locally (https://github.com/renaissance-benchmarks/renaissance/blob/510b3b9e8f01dc397f35b64a7b5f8a943b75d012/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala#L61) I'm not sure what we're trying to do with the worker instance count ?

lbulej:
I was under the impression that the number of executors influenced the default number of partitions created by default when creating RDDs directly from files on disk. It was also one of the reasons I avoided creating RDDs by just preparing data collections in memory and calling `parallelize()` on them, which would avoid I/O, but use a different class underneath. Maybe it's something we should actually do (or maybe repartition the existing RDDs explicitly).

lbulej:
> ... seeing how our Spark instance is hardcoded to run locally (
> 
> https://github.com/renaissance-benchmarks/renaissance/blob/510b3b9e8f01dc397f35b64a7b5f8a943b75d012/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/SparkUtil.scala#L61
> ) I'm not sure what we're trying to do with the worker instance count ?

[...]

Issue #283, opened 2021-07-02
CI: failing MacOS runs on JDK 11


vhotspur:
It seems that starting from 6d12a20c5f9c43218a20f451882238937ac085e3 (PR #282) runs on MacOS on JDK 11 are failing with stack overflow error.

It has happened in `als` in both builds so far ([1](https://github.com/renaissance-benchmarks/renaissance/runs/2963302138#step:10:347), [2](https://github.com/renaissance-benchmarks/renaissance/runs/2970810564#step:10:348)).

It might still be some kind of spurious error on behalf of the CI but maybe not.

```
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
NOTE: 'als' benchmark uses Spark local executor with 3 (out of 3) threads.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.apache.spark.util.SizeEstimator$ (file:/Users/runner/work/renaissance/renaissance/harness-091247-14370523251520266393/apache-spark/lib/spark-core_2.12-3.1.2.jar) to field java.util.concurrent.ConcurrentHashMap.table
WARNING: Please consider reporting this to the maintainers of org.apache.spark.util.SizeEstimator$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
====== als (apache-spark) [test], iteration 0 started ======
GC before operation: completed in 26.263 ms, heap usage 27.547 MB -> 20.566 MB.
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== als (apache-spark) [test], iteration 0 completed (4272.548 ms) ======
21/07/02 09:12:57 ERROR Executor: Exception in task 0.0 in stage 125.0 (TID 82)
java.lang.StackOverflowError
	at java.base/java.nio.ByteBuffer.position(ByteBuffer.java:1094)
	at java.base/java.nio.HeapByteBuffer.get(HeapByteBuffer.java:184)
	at org.apache.spark.util.ByteBufferInputStream.read(ByteBufferInputStream.scala:49)
	at java.base/java.io.ObjectInputStream$PeekInputStream.read(ObjectInputStream.java:2854)
	at java.base/java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2870)
[...]

PR #284, opened 2021-07-02
Set benchmark-specific thread limits for the Spark local executor

lbulej:
The limits reflect the recommendations from #260, with the limits for `gauss-mix`, `chi-square` and `dec-tree` bumped up by 2 threads (providing a bit of slack that does not help but should not hurt), resulting in the following defaults for the `spark_thread_limit` parameter:

- 4 in `chi-square` and `gauss-mix`
- 6 in `dec-tree`
- 8 in `movie-lens`
- 12 in `als`, `als-ml`, `log-regression` and `page-rank`
- unlimited in `naive-bayes`

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
@@ -24,7 +24,7 @@ import scala.util.Random
 @Repetitions(30)
 @Parameter(
   name = "spark_thread_limit",
-  defaultValue = "4",
+  defaultValue = "12",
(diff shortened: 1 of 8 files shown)


vhotspur:
Related to #283: perhaps there should be some hard minimum too?

lbulej:
Spark should be fine even with 1 CPU. On my machine, Spark benchmarks run (albeit slowly) under `taskset -c 0`, both on JDK8 and JDK11.

lbulej:
If the thread limits are fine, I'd suggest merging this, because the issue appears to be related to thread stack size. If we end up imposing a minimal number of threads to work around the thread stack size, it would be better to do it in another PR anyway.

farquet (approved):

Issue #192, opened 2019-08-09
Warning messages sometimes wrap together with the iteration line


axel22:
Suggestion is to pack the warning message into single line.

```
====== db-shootout (database), iteration 0 started ======
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== db-shootout (database), iteration 0 completed (5031.998 ms) ======
====== db-shootout (database), iteration 1 started ======
WARNING: This benchmark provides no result that can be validated.
====== db-shootout (database), iteration 1 completed (4266.008 ms) ======         There is no way to check that no silent failure occurred.

====== db-shootout (database), iteration 2 started ======
WARNING: This benchmark provides no result that can be validated.
====== db-shootout (database), iteration 2 completed (3842.75 ms) ======         There is no way to check that no silent failure occurred.

====== db-shootout (database), iteration 3 started ======
WARNING: This benchmark provides no result that can be validated.
====== db-shootout (database), iteration 3 completed (3977.732 ms) ======         There is no way to check that no silent failure occurred.

====== db-shootout (database), iteration 4 started ======
WARNING: This benchmark provides no result that can be validated.
         There is no way to check that no silent failure occurred.
====== db-shootout (database), iteration 4 completed (3969.249 ms) ======
====== db-shootout (database), iteration 5 started ======
```

PR #286, opened 2021-07-02
Print a message if the benchmark operation fails

lbulej:
This gives us a chance to at least print the name of the failure exception, because the exception can be lost if any of the tear-down methods in the *finally* blocks throws an exception.

This can actually occur in some of the Spark benchmarks because they try to dump the result of the last operation but do not expect the very first operation to fail. This assumption has been proven wrong #283 and losing the original exception makes things a bit more confusing. A fix for that is in #288.

An alternative would be to complicate the exception handling to make sure the original exception does not get lost even though the benchmark is already doomed at that point.


lbulej:
The other commit could help with #192.

farquet (approved):

PR #288, opened 2021-07-03
Add a missing null check and make dumping of Spark benchmark results optional (and disabled by default)

lbulej:
Most (but not all) Spark benchmarks dump the result of the last benchmark operation to disk during benchmark tear-down. This is primarily useful for development, but it's not necessarily something we should do all the time (or by default).

In addition, the current implementation throws a `NullPointerException` in the tear-down phase if the benchmark fails during the first operation (because the operation result will be uninitialized), which ends up masking the exception thrown from the failed benchmark operation.


farquet (approved):
Good fix. I've hit that issue myself in the past too.

lbulej:
Interestingly, not dumping the results seems to relieve the pressure on the thread stack size observed in #283, probably because we do not ask Spark to coalesce the result into a single partition.

PR #289, opened 2021-07-04
Move to ml-based variant of 'als'

lbulej:
This replaces the `mllib`-based version of `als` with the `ml`-based version. As discussed in #274, the `mllib`-based version is a wrapper around the implementation from the `ml` package. Internally, the implementation still uses RDD for the computation, which requires the `ml`-based ALS class to perform input data conversion from a DataFrame to an RDD, but the `mllib`-based ALS  class actually does the same, except from an RDD (even correctly typed) to an RDD. Because the ALS algorithm is also used by `movie-lens` which still uses the `mllib`-based ALS class, I think using the `ml`-based ALS class in `als` provides a little bit more diversity because of the different conversion.

Opinions are welcome, especially if you feel strongly about keeping the `mllib`-based version of `als`. We can pick any of the versions here (and update the PR description :-), but we should pick one.


ceresek (approved):

axel22:
From a compiler-developer perspective, there is merit in having a workload that does the wrapping and conversion, even if it is similar to another workload (as Francois pointed out, we have a similar situation with `mnemonics` and `par-mnemonics`, where even though the workload is the same, the cross-thread communication between the worker is what makes the optimizations more challenging).

lbulej:
We do have that -- `movie-lens` uses the same algorithm through the `mllib.ALS` class, so it's not like we lost something.

lbulej:
Also, the situation with `mnemonics` and `par-mnemonics` (sequential vs. parallel execution) is not entirely comparable, because the difference actually affects the entire execution. Here it only concerns input conversion but the rest of the execution is the same, which is why I don't see much value in keeping both versions of `als`, especially if we still have a benchmark that uses the older API. I would much rather introduce some entirely new workload.

Issue #241, opened 2021-04-18
Compatibility with Java 16


ceresek:
Just a quick list of benchmarks failing on Java 16 (using the 0.11.0 build), since I was running it recently:

- als
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @681a8b4e`
- chi-square
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @681a8b4e`
- db-shootout
    `java.lang.IllegalAccessError: class net.openhft.chronicle.core.OS (in unnamed module @0x5082d622) cannot access class sun.nio.ch.FileChannelImpl (in module java.base) because module java.base does not export sun.nio.ch to unnamed module @0x5082d622`
- dec-tree
    `Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end 3, length 2`
- gauss-mix
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @681a8b4e`
- log-regression
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @4bff64c2`
- movie-lens
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @4c2cc639`
- naive-bayes
    `java.lang.reflect.InaccessibleObjectException: Unable to make field transient java.lang.Object[] java.util.ArrayList.elementData accessible: module java.base does not "opens java.util" to unnamed module @681a8b4e`
- neo4j-analytics
    `Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private java.lang.String java.lang.Throwable.detailMessage accessible: module java.base does not "opens java.lang" to unnamed module @4e31276e`
- page-rank
    `Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field final byte[] java.nio.ByteBuffer.hb accessible: module java.base does not "opens java.nio" to unnamed module @5c08c46a`

[...]

PR #291, opened 2021-07-22
CI: test on more JDKs

vhotspur:
Added more environments to be tested on GitHub Actions CI.

It now runs on

 * Linux: all JDKs from 8 to 16
 * Linux with OpenJ9: 8, 11, 16
 * MacOS: 8, 11, 13, 15
 * Windows: 8, 11, 13, 15

Also added `--add-opens` to JMH runs for JDK 16 (see #241).

### DIFF
--- a/tools/ci/bench-jmh.sh
+++ b/tools/ci/bench-jmh.sh
@@ -6,6 +6,7 @@ source "$(dirname "$0")/common.sh"
 
 java -jar "$RENAISSANCE_JMH_JAR" \
 	-jvmArgs -Xms2500M -jvmArgs -Xmx2500m \
+	$( for arg in $( get_jvm_workaround_args ); do echo "-jvmArgs" "$arg"; done ) \
 	-jvmArgs -Dorg.renaissance.jmh.configuration=test \
 	-jvmArgs -Dorg.renaissance.jmh.fakeIncompatible=true \
 	-wi 0 -i 1 -f 1 -foe true
--- a/tools/ci/common.sh
+++ b/tools/ci/common.sh
@@ -21,6 +21,12 @@ RENAISSANCE_GIT_VERSION=$(git describe --tags --always --dirty=-SNAPSHOT || echo
 # Strip leading 'v' from the git-produced version
 RENAISSANCE_VERSION=${RENAISSANCE_GIT_VERSION#v}
 
+# Try to guess JVM version
+RENAISSANCE_JVM_MAJOR_VERSION="$( java -XshowSettings:properties -version 2>&1 \
+    | sed -n -e 's#^[ \t]*java.version[ ]*=[ ]\(.*\)#\1#p' \
+    | sed -e 's#1\.8#8#' -e 's#\([^.]*\).*#\1#' \
+    || echo 8 )"
+
 # The base bundle
 RENAISSANCE_DIR="target"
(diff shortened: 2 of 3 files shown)


tools/ci/common.sh
```diff
@@ -21,6 +21,12 @@ RENAISSANCE_GIT_VERSION=$(git describe --tags --always --dirty=-SNAPSHOT || echo
 # Strip leading 'v' from the git-produced version
 RENAISSANCE_VERSION=${RENAISSANCE_GIT_VERSION#v}
 
+# Try to guess JVM version
+RENAISSANCE_JVM_MAJOR_VERSION="$( java -XshowSettings:properties -version 2>&1 \
+    | sed -n -e 's#^[ \t]*java.version[ ]*=[ ]\(.*\)#\1#p' \
```
lbulej, line 26: I would write the `sed` expression as follows:
```
sed -n '/^[[:blank:]]*java[.]version[[:blank:]]*=/ { s/.*=[[:blank:]]*\([^[:space:]]\+\)/\1/; s/^1[.]//; s/^\([^.]\+\).*/\1/; p }'
```
but it's your call. I'm a bit uncomfortable with the `#` characters in the `sed` expression (my brain sees comments).

Also, is it necessary to use the `-X` option? The following works with just the `-version` output on JDK 8, 11, 16, and Hotspot 15. On OpenJ9 11, it gets the opendjk version from the first line, which is probably fine too:
```
sed -n '/version[[:blank:]]\+/ { s/.*version[[:blank:]]\+"\([^[:space:]"]\+\)/\1/; s/^1[.]//; s/^\([^.]\+\).*/\1/; p }'
```

If we really want to use the `-X` option, why not ask for the `java.specification.version`? The `sed` expression only needs to strip `1.` from the `1.8` version (you could even keep it there).
```
 sed -n '/^[[:blank:]]*java[.]specification[.]version[[:blank:]]*=/ { s/.*=[[:blank:]]*\([^[:space:]]\+\)/\1/; s/^1[.]//; p }'
```
This worked on OpenJDK 8, 11, 16, Hotspot 15, and OpenJ9 11.

tools/ci/bench-jmh.sh
```diff
@@ -6,6 +6,7 @@ source "$(dirname "$0")/common.sh"
 
 java -jar "$RENAISSANCE_JMH_JAR" \
 	-jvmArgs -Xms2500M -jvmArgs -Xmx2500m \
+	$( for arg in $( get_jvm_workaround_args ); do echo "-jvmArgs" "$arg"; done ) \
```
lbulej, line 9: Unless we put other workaround args in here, I'd suggest to just name it `get_jvm_add_opens_args` for now.

lbulej (approved):
LGTM

tools/ci/common.sh
vhotspur: Okay, using `java -version` (hopefully they will not unify it with `--version` that has slightly different output) and used `/` as delimiter in `sed` ;-) [tested on JDK 8 to 16 including OpenJ9].

tools/ci/bench-jmh.sh
vhotspur: I would say that the fact that it adds only `--add-opens` is an implementation detail from the point of view of the wrapper script.

lbulej:
LEBTM :-)

ceresek:
That's just one step away from LoBoToMy :-)

PR #292, opened 2021-07-22
Specify plugin class inside plugin JAR

### LINKED ISSUE #194: Specify plugin class inside plugin JAR

vhotspur:
Add `Renaissance-Plugin` property to a `MANIFEST.MF` of a plugin describing plugin main class. When the plugin is loaded, the class is automatically read from the manifest. Simplifies the command line, fixes #194.

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
+++ b/renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
@@ -321,21 +321,33 @@ public final class ModuleLoader {
   static <T> Class<? extends T> loadExtension(
     Collection<Path> classPath, String className, Class<T> baseClass
   ) throws ModuleLoadingException {
-    URL[] classPathUrls = pathsToUrls(classPath);
-    if (logger.isLoggable(Level.CONFIG)) {
-      logger.config(String.format(
-        "Class path for %s: %s", className,
-        Arrays.stream(classPathUrls).map(Object::toString).collect(joining(","))
-      ));
+    ClassLoader loader = createClassLoaderFromPaths(classPath, className);
+    return loadExtensionFromClassLoader(loader, className, baseClass);
+  }
+
+
+  /**
+   * Loads an extension class and casts it to the desired base class.
+   * The class name is read from a given property.
+   */
+  static <T> Class<? extends T> loadDescribedExtension(
+    Collection<Path> classPath, String propertyName, Class<T> baseClass
+  ) throws ModuleLoadingException {
+    ClassLoader loader = createClassLoaderFromPaths(classPath, propertyName);
+    String className = getManifestProperty(loader, propertyName);
+
+    if (className == null) {
+      throw new ModuleLoadingException("classname to load not found in manifests");
     }
 
-    if (classPathUrls.length != classPath.size()) {
-      throw new ModuleLoadingException("malformed URL(s) in classpath specification");
-    }
+    return loadExtensionFromClassLoader(loader, className, baseClass);
+  }
 
+  /** Loads extension from initialized class loader. */
+  private static <T> Class<? extends T> loadExtensionFromClassLoader(
+    ClassLoader loader, String className, Class<T> baseClass
+  ) throws ModuleLoadingException {
     try {
-      ClassLoader parent = ModuleLoader.class.getClassLoader();
-      ClassLoader loader = new URLClassLoader(classPathUrls, parent);
       Class<?> loadedClass = loader.loadClass(className);
       return loadedClass.asSubclass(baseClass);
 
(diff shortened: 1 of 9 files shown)


renaissance-harness/src/main/scala/org/renaissance/harness/ConfigParser.scala
```diff
@@ -51,11 +51,6 @@ private final class ConfigParser(tags: Map[String, String]) {
         .valueName("<class-path>!<class-name>")
         .text("Load external plugin. Can appear multiple times.")
         .action((v, c) => c.withPlugin(v))
```
lbulej, line 53: We might want to indicate that class name is optional, i.e., `<class-path>[!<class-name>]`

renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
```diff
@@ -351,6 +364,52 @@ private static void extractResource(String resourcePath, Path outputFilePath)
...
+          }
+        } catch (IOException e) {
+          continue;
+        }
+      }
+    } catch (IOException e) {
+      // Ignore.
+    }
+    return null;
+  }
```
lbulej, line 412: I would move it to the `BenchmarkSuite` class, it only supports the management decision that is being taken in the `BenchmarkSuite`. I plan to move (static) methods methods that are not essential for the module loading mechanism from the `ModuleLoader` class into the `BenchmarkSuite` class. It's not something critical, but at least in this case, I'd avoid putting more code there.

lbulej (approved):
LGTM in general. Do we need a README update?

renaissance-harness/src/main/scala/org/renaissance/harness/ConfigParser.scala
vhotspur: Done, thanks!

renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
vhotspur: > I would move it to the `BenchmarkSuite` class, it only supports the management decision that is being taken in the `BenchmarkSuite`.

That makes sense. However, it is used internally by the `loadDescribedExtension` as I wanted to reuse the initialized class-loader. I think that moving just this one method would not make sense without further changes. But I do not see what to move in this case unless we want to duplicate the loader initialization. Any ideas?

lbulej: > That makes sense. However, it is used internally by the `loadDescribedExtension` as I wanted to reuse the initialized class-loader.

I overlooked that bit.

> I think that moving just this one method would not make sense without further changes. But I do not see what to move in this case unless we want to duplicate the loader initialization. Any ideas?

Actually, because "extension" is a concept of the suite, I really wanted to pull all extension loading out of `ModuleLoader` and have it only provide lower-level services related to loading a class using a classloader built from a bunch of jars (which sometimes need to be unpacked) so that it is useful both to `Launcher` and `BenchmarkSuite` classes. The responsibility of the `Launcher` and the `BenchmarkSuite` is to determine the module (or the jars that make up the class path) and the class name, and ask the `ModuleLoader` to load (and possibly instantiate) the class.

Can you just move a bigger chunk, i.e., the `loadDescribedExtension` as well? The rest can come later. But if you don't feel like doing it in this change set, we can leave it for later.

renaissance-harness/src/main/scala/org/renaissance/harness/MarkdownGenerator.scala
```diff
@@ -374,8 +374,9 @@ operation, and will pass the result of the `isLast` method to some other events.
 
 To make the harness use an external plugin, it needs to be specified on the command line.
 The harness can load multiple plugins, and each must be enabled using the
-`--plugin <class-path>!<class-name>` option. The `<class-path>` is the class path on which
-to look for the plugin class, and `<class-name>` is a fully qualified name of the plugin class.
+`--plugin <class-path>[!<class-name>]` option. The `<class-path>` is the class path on which
+to look for the plugin class, and `<class-name>` is a fully qualified name of the plugin class
+(the class-name can be omitted when specified inside the plugin JAR in a `MANIFEST.MF` property).
```
lbulej, line 379: Sorry to nit-pick, can you sneak the name of the property in here, i.e., that it is the `Renaissance-Plugin` property?

vhotspur: Done.

renaissance-core/src/main/java/org/renaissance/core/ModuleLoader.java
vhotspur: > Can you just move a bigger chunk, i.e., the `loadDescribedExtension` as well? The rest can come later. But if you don't feel like doing it in this change set, we can leave it for later.

That should include moving the plain `loadExtension` as well and perhaps something else too. I will do that in a separate PR.

lbulej: OK

PR #293, opened 2021-07-23
Module loading refactoring

vhotspur:
A follow up to the discussion in #292.

@lbulej Have you meant something like this?

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/core/BenchmarkSuite.java
+++ b/renaissance-core/src/main/java/org/renaissance/core/BenchmarkSuite.java
@@ -205,7 +205,17 @@ public final class BenchmarkSuite {
 
   // Extension support
 
-/** Read all manifests and find first one having given property.
+  public static final class ExtensionException extends Exception {
+    ExtensionException(String message) {
+      super(message);
+    }
+
+    ExtensionException(String format, Object... args) {
+      super(String.format(format, args));
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


lbulej:
> A follow up to the discussion in #292.
> 
> @lbulej Have you meant something like this?

Almost :-) But we are splitting hairs here, so take it with a grain of salt.

I would have skipped the extra step with `ExtensionUtils` and moved the code directly into `BenchmarkSuite`.
The contents of `ExtensionUtils.create` would become `BenchmarkSuite.createExtension` (and similarly, the contents of `ExtensionUtils.createPropertyDescribed` would become `BenchmarkSuite.createDescribedExtension`).

The `getManifestProperty` method would go to `BenchmarkSuite`, but the remaining two static methods might need a bit more thought. I was not comfortable with throwing the `ModuleLoadingException` (an inner class intended to be thrown only by the `ModuleLoader`) from another class, which is why I did not move things to the `BenchmarkSuite` class immediately -- at that point, I was happy to make `ModuleLoader` package private (almost, it's still public because of the bloody `MarkdownGenerator`) and avoid duplicating code in the harness and in the JMH wrapper :-)

vhotspur:
@lbulej Do you think this can be merged in its current state?

lbulej (approved):
I can't focus on this yet, so I would say go ahead. This does not change the workloads in any way, so I believe it is safe.

PR #294, opened 2021-07-23
Documentation improvements

vhotspur:
Various bits of missing documentation (not ready for merging, opening PR to track progress).

I was also wondering whether not to split a bit more README.md and CONTRIBUTING.md. The README now is quite long and mixes things for users as well as for developers. I would rather have the README shorter with instructions on how to run it (assuming the user downloaded a release) and keep separate documents for building and about the design.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,27 +14,14 @@ This will retrieve all the dependencies, compile all the benchmark projects and 
 bundle the JARs and create the final JAR under the `target` directory.
 
 
-### Code organization and internals
-
-The code is organized into three main parts:
-
-- `renaissance-core`: these are the core APIs that a benchmark needs to work with,
-  such as the runtime configuration, the benchmark base class or the policy.
-- `renaissance-harness`: this is the overall suite project, which is responsible for
-  parsing the arguments, loading the classes, and running the benchmark.
-- a set of projects in the `benchmarks` directory: these are the individual groups of benchmarks
-  that share a common set of dependencies. A group is typically thematic, relating to
-  a particular framework or a JVM language, and benchmarks in different groups depend
-  on a different set of libraries, or even the same libraries with different versions.
-  The bundling mechanism of the suite takes care that the dependencies of the different groups
-  never clash with each other.
-
-
 ### Adding a new benchmark
 
 To add a new benchmark to an existing group, identify the respective project
 in the `benchmarks` directory, and add a new top-level Scala (Java) class
 that extends (implements) the `Benchmark` interface.
+
+We recommend you also consult [Design overview document](documentation/design.md)
+to understand how the whole project is organized.
 
 Here is an example:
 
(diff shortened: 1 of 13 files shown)


ceresek:
+1 for separating the documentation about development and contributing :-)

lbulej (approved):
LGTM.

Also, I'm all for making `README.md` simpler.

Even more, I would like to get rid of `MarkdownGenerator` from the harness project. If we don't start using some kind of templating engine for generating the markdown with implementation details, we could try to reuse the `MarkdownGenerator`  as a resource generator in SBT. It sets up a dictionary of symbols during initalization, so it might be easy to switch to loading a property files with symbols that we either maintain by hand, or we produce from code (initially by reusing the initialization part of `MarkdownGenerator`).

ceresek:
The only thing I would be worried about when using an external templating engine is that it represents another dependency that we need to remain compatible across a wide range of JDK versions. Other than that, things like Thymeleaf look good ...

lbulej:
> The only thing I would be worried about when using an external templating engine is that it represents another dependency that we need to remain compatible across a wide range of JDK versions. Other than that, things like Thymeleaf look good ...

That's actually my concern as well. It's just two files, the Scala string interpolation works reasonably well for this, and I think the hassle with integrating a templating engine into the build system is not worth it, because I cannot estimate anything with SBT. As far as I'm concerned, my goal was to first move the bloody `MarkdownGenerator` from the harness project into a build project and stop special casing the `--readme` option in the `Launcher`.

The next step was to come up with a way to provide the generator with the input it needs without it forcing package private classes to be public. But apart from a hand-maintained property file, the only other way I came up with was generating the relevant parts of the property file each subproject and then merge them before feeding it to the generator.

But orchestrating this in SBT would mean a lot of trial and error (at least for me), in part because we inflict more SBT pain on ourselves than necessary (IMO), because we work with subprojects in a way that optimizes for the wrong case and goes against SBT design intentions. And fixing that is something that eventually want to do, but it's in the category of trial-and-error work with uncertain benefits -- it might also just end up proving me wrong, so I don't really rush into that.

lbulej:
I know we haven't got rid of the `MarkdownGenerator` yet, but perhaps this could go in as well? It should be innocent enough.

vhotspur:
> I know we haven't got rid of the `MarkdownGenerator` yet, but perhaps this could go in as well? It should be innocent enough.

I am not finished with this yet (I hope I will be able to finish it later today) but it shall be in a mergeable state.

vhotspur:
> I know we haven't got rid of the `MarkdownGenerator` yet, but perhaps this could go in as well? It should be innocent enough.

@lbulej I did some more splitting and I have (hopefully) streamlined the contribution guide a bit. There is never enough documentation but I think this can be merged. Please, feel free to merge before the release if the result looks sane to you :-). Thanks.

lbulej:
Looks perfect to me. Thanks for this (usually thank-less) work!

PR #297, opened 2021-09-15
Update from scala 2.12.14 to 2.12.15 for complete JDK17 compatibility

farquet:
Scala 2.12.15 landed -> https://github.com/scala/scala-dev/issues/783

So we can bump our 2.12.14 dependencies and do a renaissance 0.13.0 release which will officially support the now released JDK17.

### DIFF
--- a/benchmarks/actors-reactors/build.sbt
+++ b/benchmarks/actors-reactors/build.sbt
@@ -8,7 +8,7 @@ lazy val actorsReactors = (project in file("."))
     name := "actors-reactors",
     version := (renaissanceCore / version).value,
     organization := (renaissanceCore / organization).value,
-    scalaVersion := "2.12.14"
+    scalaVersion := "2.12.15"
   )
   .dependsOn(
     renaissanceCore % "provided",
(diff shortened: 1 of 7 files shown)


lbulej (approved):
LGTM.

When I package the release, do we want to call it 0.12.1 or 0.13.0? I can see using either of the two (i.e., 0.12.1 indicates minor change in the suite, no changes in benchmark code vs. 0.13 is **the version** that supports JK8, JDK11 and JDK17).

ceresek:
Let's do 0.13 :) we have to tempt the luck :)

farquet:
@lbulej it's a good point that the benchmarks haven't changed. Moreover, the scala update is a minor compatibility one. So performance-wise this will be the same. I would then tend to go for 0.12.1 then. But no strong opinion on that. Especially since I don't want to contradict @ceresek :) 

@vhotspur, I've checked how to add JDK17 CI testing, but it seems that it will require some changes there -> https://github.com/renaissance-benchmarks/docker-buildenv
Not sure how to do it. Would you mind looking into it?

Btw, if resources allow, no problem keep JDK13 and JDK15 testing, but I guess the only thing that really matters these days is to stay compatible with the LTS versions, so that would be JDK8, JDK11, and now JDK17. With the release cadence down to every 2 years for LTS as announced yesterday, it seems that only testing LTS versions is reasonable.

PR #298, opened 2021-09-15
CI: test on JDK17

vhotspur:
Follow-up to PR #297.

JDK17 for Mac and Windows is not yet available, tracked here: actions/setup-java#229

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,7 @@ jobs:
           - openjdk14
           - openjdk15
           - openjdk16
+          - openjdk17
           - openj9-openjdk8
           - openj9-openjdk11
           - openj9-openjdk16
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,7 +106,7 @@ jobs:
   macos:
     strategy:
       matrix:
-        java: [ '8', '11', '13', '15' ]
+        java: [ '8', '11', '13', '15', '17' ]
     runs-on: macos-latest
     continue-on-error: true
     steps:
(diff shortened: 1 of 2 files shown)


lbulej:
> JDK17 for Mac and Windows is not yet available, tracked here: [actions/setup-java#229](https://github.com/actions/setup-java/issues/229)

By "not available", you mean just on GHA, right?

Given that JDK17 is available as a download, I'd like to see the 0.13 release running on Mac and Windows in addition to Linux. Do we wait for GHA to support it out of the box, or we try to install it ourselves now and simplify the CI after GHA starts supporting JDK17?

vhotspur:
> > JDK17 for Mac and Windows is not yet available, tracked here: [actions/setup-java#229](https://github.com/actions/setup-java/issues/229)
> 
> By "not available", you mean just on GHA, right?

Yes.

> Given that JDK17 is available as a download, I'd like to see the 0.13 release running on Mac and Windows in addition to Linux. Do we wait for GHA to support it out of the box, or we try to install it ourselves now and simplify the CI after GHA starts supporting JDK17?

I would prefer to wait (I have nowhere to reasonably test it) but I guess we can use [this workaround](https://github.com/actions/setup-java/issues/229#issuecomment-919678258) if you need it for 0.13. I will have a look.

BTW, when you are in process of making a release, would you, please, add some kind of bullet list of things that needs to be done to the documentation branch (PR #294)?

farquet:
I've manually tested on Mac and it works just fine. Linux is tested in the CI. So if someone can give it a shot on Windows just to be sure, I think we're safe for the release.

Regarding the CI, we can just keep this PR open until JDK17 is available on GHA for Mac and Windows. Or do a subsequent PR later on, but in any case, it shouldn't block the release. We have to take advantage of the JDK17 release momentum to do the renaissance release :)

vhotspur:
Summary: testing with JDK17 on Mac, Windows and Linux, can be merged (the failed checks are spurious problems IMHO).

> I've manually tested on Mac and it works just fine. Linux is tested in the CI. So if someone can give it a shot on Windows just to be sure, I think we're safe for the release.

After figuring out how to use `aria2` adding Mac was relatively easy (I just had to fix `sed` portability issues (again)).

> Regarding the CI, we can just keep this PR open until JDK17 is available on GHA for Mac and Windows. Or do a subsequent PR later on, but in any case, it shouldn't block the release. We have to take advantage of the JDK17 release momentum to do the renaissance release :)

After release, we shall re-think how CI should work because this is becoming unmanageable (too many jobs, impossible to restart just one, always testing platforms that are not that interesting, running jobs twice for PRs etc.).

farquet (approved):

PR #299, opened 2021-09-16
Set length of abbreviated commit id in git describe

lbulej:
This seems to be needed because the command (as used in various scripts) suddenly started returning commit id with one extra
character, resulting in a mismatch between `git describe` and the git-described version produced by the `sbt-git` plugin (at least on my machine with git 2.31.1). Due to this, the pre-push readme check fails to find the right JAR and aborts the push.

### DIFF
--- a/tools/ci/common.sh
+++ b/tools/ci/common.sh
@@ -16,7 +16,7 @@ ROOT_DIR="$(git rev-parse --show-toplevel || realpath "$( dirname "$0" )/../../"
 CACHE_DIR="$HOME/.prebuilt"
 
 # Raw version as produced by git
-RENAISSANCE_GIT_VERSION=$(git describe --tags --always --dirty=-SNAPSHOT || echo "ci-SNAPSHOT" )
+RENAISSANCE_GIT_VERSION=$(git describe --tags --always --abbrev=7 --dirty=-SNAPSHOT || echo "ci-SNAPSHOT" )
 
 # Strip leading 'v' from the git-produced version
 RENAISSANCE_VERSION=${RENAISSANCE_GIT_VERSION#v}
--- a/tools/pre-push
+++ b/tools/pre-push
@@ -36,7 +36,7 @@ echo ""
 # Check that README.md is up-to-date.
 echo ":: Running pre-push markdown check ::"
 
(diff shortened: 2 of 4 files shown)


farquet (approved):

vhotspur:
Perhaps it would make sense to ensure (enforce) that the same length is always used by sbt-git too (and upgrading the length to 12 to make it last virtually forever).

```scala
git.useGitDescribe := false

git.formattedShaVersion := {
  val describedVersion = git.gitDescribedVersion.value.getOrElse("")
  val describedVersionNoHash = describedVersion.substring(0, describedVersion.lastIndexOf("-"))
  val shortCommitSha = git.gitHeadCommit.value.getOrElse("").substring(0, 7)
  val suffix = git.makeUncommittedSignifierSuffix(git.gitUncommittedChanges.value, git.uncommittedSignifier.value)
  git.gitHeadCommit.value map { _ => s"${describedVersionNoHash}-g${shortCommitSha}${suffix}" }
}
```

lbulej:
> Perhaps it would make sense to ensure (enforce) that the same length is always used by sbt-git too (and upgrading the length to 12 to make it last virtually forever).

It surely would, but looking into `sbt-git` code did not reveal an `abbrev`-like setting that I could use easily, without having to experiment with SBT tasks :-) What you propose looks fine and can go right in even though I would like it better as a `sbt-git` setting.

vhotspur:
> I would like it better as a sbt-git setting.

Same here. I only found sbt/sbt-git#155 (from 2018) and the code above was a mix of existing sbt-git source code and other fragments from GitHub. It looks ugly but I doubt there is a better way (because `describedVersion` is never constructed from parts as it comes from the command itself or from JGit).

lbulej:
> > I would like it better as a sbt-git setting.
> 
>  ... (because describedVersion is never constructed from parts as it comes from the command itself or from JGit).

This is why I did not originally understand why the pre-commit hook was failing. I always assumed that `sbt-git` just gives me the output of `git describe`, so even if `git` suddenly started using more characters, `sbt-git` should just pass it on. It took me a while to discover it, because the difference in one character in the JAR file name was barely discernible and I saw what I expected to see :-) I didn't find anything immediately amiss in `sbt-git` and did not dig into JGit.

PR #300, opened 2021-09-16
Add release notes for the upcoming v0.13.0

lbulej:
There is also a draft of GitHub release (shorter) that you should be able to see. If we are fine with this, I'll go ahead with the release.
Note: this is built on top of #299 so that I can push to GitHub.


farquet (approved):
awesome !

PR #301, opened 2021-09-16
Add a description of the release process

lbulej:
... so that people don't have to re-discover it on their own with every release...


vhotspur (approved):
Detailed and perfect, thanks a lot :-)

PR #302, opened 2021-09-30
Make system info dump more resilient

ceresek:
Substrate VM does not support some of the system info queries we do when dumping JSON.

```
SEVERE: Harness failed with com.oracle.svm.core.jdk.UnsupportedFeatureError: OperatingSystemMXBean methods
	at com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:87)
	at com.oracle.svm.core.jdk.management.SubstrateOperatingSystemMXBean.getFreePhysicalMemorySize(SubstrateOperatingSystemMXBean.java:107)
	at org.renaissance.harness.JsonWriter.getOsInfo(ResultWriter.scala:184)
	at org.renaissance.harness.JsonWriter.getEnvironment(ResultWriter.scala:164)
	at org.renaissance.harness.JsonWriter.store(ResultWriter.scala:363)
	at org.renaissance.harness.ResultWriter.org$renaissance$harness$ResultWriter$$storeResults(ResultWriter.scala:63)
	at org.renaissance.harness.ResultWriter.beforeHarnessShutdown(ResultWriter.scala:67)
	at org.renaissance.harness.EventDispatcher.notifyBeforeHarnessShutdown(EventDispatcher.java:76)
	at org.renaissance.harness.RenaissanceSuite$.runBenchmarks(RenaissanceSuite.scala:163)
	at org.renaissance.harness.RenaissanceSuite$.main(RenaissanceSuite.scala:117)
	at java.lang.reflect.Method.invoke(Method.java:566)
	at org.renaissance.core.Launcher.loadAndInvokeHarnessClass(Launcher.java:114)
	at org.renaissance.core.Launcher.launchHarnessClass(Launcher.java:73)
	at org.renaissance.core.Launcher.main(Launcher.java:37)
```

The runtime can return some paths as NULL.

```
SEVERE: Harness failed with java.lang.NullPointerException
	at org.renaissance.harness.JsonWriter.pathsAsJson(ResultWriter.scala:161)
	at org.renaissance.harness.JsonWriter.getJreInfo(ResultWriter.scala:241)
	at org.renaissance.harness.JsonWriter.getEnvironment(ResultWriter.scala:167)
	at org.renaissance.harness.JsonWriter.store(ResultWriter.scala:367)
	at org.renaissance.harness.ResultWriter.org$renaissance$harness$ResultWriter$$storeResults(ResultWriter.scala:64)
	at org.renaissance.harness.ResultWriter.beforeHarnessShutdown(ResultWriter.scala:68)
	at org.renaissance.harness.EventDispatcher.notifyBeforeHarnessShutdown(EventDispatcher.java:76)
	at org.renaissance.harness.RenaissanceSuite$.runBenchmarks(RenaissanceSuite.scala:163)
	at org.renaissance.harness.RenaissanceSuite$.main(RenaissanceSuite.scala:117)
	at java.lang.reflect.Method.invoke(Method.java:566)
	at org.renaissance.core.Launcher.loadAndInvokeHarnessClass(Launcher.java:114)
	at org.renaissance.core.Launcher.launchHarnessClass(Launcher.java:73)
	at org.renaissance.core.Launcher.main(Launcher.java:37)
```

### DIFF
--- a/renaissance-harness/src/main/scala/org/renaissance/harness/ResultWriter.scala
+++ b/renaissance-harness/src/main/scala/org/renaissance/harness/ResultWriter.scala
@@ -187,16 +187,14 @@ private final class JsonWriter(val jsonFile: Path) extends ResultWriter {
     os match {
       case unixOs: UnixOperatingSystemMXBean =>
         // Gag possible exceptions.
-        Try(
-          result ++= Seq(
-            "phys_mem_total" -> unixOs.getTotalPhysicalMemorySize.toJson,
-            "phys_mem_free" -> unixOs.getFreePhysicalMemorySize.toJson,
-            "virt_mem_committed" -> unixOs.getCommittedVirtualMemorySize.toJson,
-            "swap_space_total" -> unixOs.getTotalSwapSpaceSize.toJson,
-            "swap_space_free" -> unixOs.getFreeSwapSpaceSize.toJson,
-            "max_fd_count" -> unixOs.getMaxFileDescriptorCount.toJson,
-            "open_fd_count" -> unixOs.getOpenFileDescriptorCount.toJson
-          )
+        result ++= Seq(
+          "phys_mem_total" -> Try(unixOs.getTotalPhysicalMemorySize).toOption.toJson,
+          "phys_mem_free" -> Try(unixOs.getFreePhysicalMemorySize).toOption.toJson,
+          "virt_mem_committed" -> Try(unixOs.getCommittedVirtualMemorySize).toOption.toJson,
+          "swap_space_total" -> Try(unixOs.getTotalSwapSpaceSize).toOption.toJson,
+          "swap_space_free" -> Try(unixOs.getFreeSwapSpaceSize).toOption.toJson,
+          "max_fd_count" -> Try(unixOs.getMaxFileDescriptorCount).toOption.toJson,
+          "open_fd_count" -> Try(unixOs.getOpenFileDescriptorCount).toOption.toJson
         )
 
       // No extra information to collect on non-Unix systems.
--- a/renaissance-harness/src/main/scala/org/renaissance/harness/ResultWriter.scala
+++ b/renaissance-harness/src/main/scala/org/renaissance/harness/ResultWriter.scala
@@ -16,6 +16,7 @@ import java.nio.file.StandardOpenOption
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.TimeZone
+import scala.util.Try
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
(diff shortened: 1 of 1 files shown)


farquet:
Those features are indeed still missing on the native image side: https://github.com/oracle/graal/blob/vm-21.2.0/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/jdk/management/SubstrateOperatingSystemMXBean.java#L73-L120

vhotspur (approved):
LGTM.

lbulej:
I'll probably get laughed at, but I'm a bit uncomfortable with the blanket `Try` the value of which nobody cares about and the fact that we discard everything if there is even a single exception, even though I'm fully aware that right now we don't get any info and a crash as a bonus.

@farquet Francois, are (some of) the MX beans in the native image completely unimplemented, or do they implement methods selectively? If it's the former, then we are probably fine with this fix, even though I would welcome a comment on why there might be exceptions to gag. If it's the latter (and if we want all information that we can get without crashing), then we might want to deal with the failures on a per-call basis.

lbulej:
Looking at the code Francois referenced, only the "total memory size" returns something useful. Everything else is dead and I can't guess whether all of it will be (eventually) provided at the same time, or whether it will somehow trickle in. If we don't care about the total memory size on native image, we can do the blanket fix for now and possibly remove it (the MX beans methods are not supposed to fail) if it gets fixed in the native image.

ceresek:
Would you be fine with wrapping each getter in `Try`, something like `"phys_mem_total" -> Try (unixOs.getTotalPhysicalMemorySize).getOrElse (0).toJson` ?

ceresek:
(Just not sure what value we should pick for N/A ... or maybe we should exclude the fields entirely in case of exception ?)

lbulej:
... exactly the kind of questions I was asking myself, which is why I would be OK with the blanket fix (maybe not using `Try`, which is used for chaining potentially failing computations) if this was just a native image thing that might go away sooner or later, and because I feel that it borders with nitpicking (I need to keep my OCD in check).

Fixing it per value using `Try` is fine, except I would avoid using zeroes as default values -- I think JSON supports the `null` literal, which looks as a reasonable substitute for `NA` in the data to me.

Alternatively, `Try` could be applied to the whole tuple, so that in case of an exception we skip a name/value pair. I would not go that way, because to me, the `null` value in the original data file carries a bit of extra information, i.e., the data should be there but isn't, so something is fishy. In contrast, on non-unix systems, the entries are not even in the JSON file, because they are not supported by the platform.

I don't know if that bit of information is worth anything to anyone, because in the end, when you **work** with the JSON files and start assuming that things may be missing, you need to come with sensible defaults on your end. Moreover, we probably don't really use that information at the moment, so it's just for getting a better picture of the platform we ran on. I'm just cautious because the JSON is a consumable produced by the benchmark that is being interpreted and thus becomes part of its public interface...

ceresek:
Better ? :-)

lbulej:
> Better ? :-)

Oh yes :-)

lbulej (approved):

Issue #296, opened 2021-08-12
Most/all benchmarks from 0.12.0 seem to fail on Apple Silicon with macOS Big Sur and Azul JDK 11 for AArch64


dcominottim:
Below is the terminal output I got after letting the benchmarks run overnight on my MacBook Pro M1. The JDK version used was 11.0.12+7 (Zulu: 11.50.19) for AArch64.

    danilo@MacBook-Pro-de-Danilo Downloads % java -jar renaissance-mit-0.12.0.jar all
    Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
    NOTE: 'page-rank' benchmark uses Spark local executor with 8 (out of 8) threads.
    WARNING: An illegal reflective access operation has occurred
    WARNING: Illegal reflective access by org.apache.spark.util.SizeEstimator$ (file:/Users/danilo/Downloads/harness-020944-17790934427461370637/apache-spark/lib/spark-core_2.12-3.1.2.jar) to field java.util.concurrent.ConcurrentHashMap.table
    WARNING: Please consider reporting this to the maintainers of org.apache.spark.util.SizeEstimator$
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release
    ====== page-rank (apache-spark) [default], iteration 0 started ======
    GC before operation: completed in 23.798 ms, heap usage 633.000 MB -> 21.942 MB.
    ^C====== page-rank (apache-spark) [default], iteration 0 failed (SparkException) ======
    
    danilo@MacBook-Pro-de-Danilo Downloads % java -jar renaissance-mit-0.12.0.jar all > renaissance-run1.txt
    Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
    WARNING: An illegal reflective access operation has occurred
    WARNING: Illegal reflective access by org.apache.spark.util.SizeEstimator$ (file:/Users/danilo/Downloads/harness-021021-6529297898483024289/apache-spark/lib/spark-core_2.12-3.1.2.jar) to field java.util.concurrent.ConcurrentHashMap.table
    WARNING: Please consider reporting this to the maintainers of org.apache.spark.util.SizeEstimator$
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release
    WARNING: This benchmark provides no result that can be validated.
    There is no way to check that no silent failure occurred.
[...]

Issue #153, opened 2019-06-14
db-shootout  does not work on linux ppc64le / s390x


MBaesken:
It seems the  db-shootout benchmark does not work on linux ppc64le / s390x .
It contains a native lib  (leveldbjni64-1.8 ) that is not available on all CPU/HW platforms where OpenJDK is supported.
Error message (example is from linux ppc64le) 
java.lang.UnsatisfiedLinkError: Could not load library. Reasons: [no leveldbjni64-1.8 in java.library.path, no leveldbjni-1.8 in java.library.path, no leveldbjni in java.library.path, /tmp/libleveldbjni-64-1-8490766625802501468.8: /tmp/libleveldbjni-64-1-8490766625802501468.8: cannot open shared object file: No such file or directory (Possible cause: can't load AMD 64-bit .so on a Power PC 64 LE-bit platform)]  

Probably the db-shootout benchmark should be run only on the platforms that are supposed to work (means:  has the native library available).

MBaesken:
btw. the reported issues has been seen with both JDK8 and JDK11.

farquet:
Thanks Matthias for the report.
The `leveldbjni-all` dependency is "supposed" to support all platforms, unfortunately, it doesn't seem to be the case as one can see by unzipping the jar at http://central.maven.org/maven2/org/fusesource/leveldbjni/leveldbjni-all/1.8/leveldbjni-all-1.8.jar

I see different options here : 
1. Add the missing native libs to a fork of the leveldbjni-all and push it to maven since the owners don't seem willing to do it : https://github.com/fusesource/leveldbjni/issues/85
2. Retire the benchmark since it would be convenient not to have benchmarks relying on native libs for compatibility reasons. That would require a good replacement benchmarks though.
3. Mark supported platforms for each benchmark and skip them when running on an unsupported platform.
4. Remove LevelDB from that benchmark since the benchmark consists of four different DBs tested. That probably wouldn't hurt so much to go down to three. https://github.com/renaissance-benchmarks/renaissance/blob/master/benchmarks/database/src/main/scala/org/renaissance/database/DbShootout.scala#L69-L91

Any opinion welcome @lbulej, @vhotspur, @axel22 & co.

I personally think that option 4 may be a good tradeoff. It's a simple change that would solve that particular issue. Moreover, the benchmark would probably not behave very differently. That would need to be quantify though.

axel22:
I also think that option 4 is preferable for now.

[...]

PR #322, opened 2021-10-12
Update and unify JNA

lbulej:
This was primarily triggered by the issues with `db-shootout` on less common CPU/platform combinations (see #153 and #296) and at least part of the cause is that [JNA](https://github.com/java-native-access/jna) version (4.2.1) used in `db-shootout` simply does not support those combinations. So this forces JNA in `db-shotout` to version 5.9.0 ([latest](https://github.com/java-native-access/jna/releases) at the moment) which provides support for currently live combinations and loses a few outdated ones, specifically:
- adds `aix-ppc64`, `darwin-aarch64`, `darwin-x86-64`, `linux-armel`, `linux-mips64el`, `linux-ppc` (32-bit), `linux-riscv64`, `linux-s390x`, and `win32-aarch64`
- removes `darwin` (replaced by `darwin-x86-64` I think), `linux-ppc64` (there is still `linux-ppc64le`), `linux-sparcv9` (there still is `sunos-sparcv9`) and `w32ce-arm`

Given the two issues mentioned, we are mostly interested in getting `darwin-aarch64`, `linux-ppc64le` and `linux-s390x` support which the newer JNA provides.

Also, because the `neo4j` and `scala-dotty`  benchmarks pull in JNA as a (transitive) dependency, I forced them to pull the same (latest) version so that we don't have three different JNA versions each supporting a different set of CPU/platform combinations. I don't know if these benchmarks actually get to executing any code that relies on JNA (not sure about `neo4j`, but `scala-dotty` might pull it in just because some part of Scala wants to use `jline-terminal` and `jline-terminal-jna` for REPL).

Anyway, the benchmarks compile and run on my machine (which is expected), but unfortunately I don't have info about the exotic platforms involved in the original issues.

I created a branch and a special build (including binaries) of `db-shootout` that the original issue posters could use to check if it fixes their problem, but I haven't received any response so far.

So the question is, do we want to merge this and move on?

### DIFF
--- a/benchmarks/database/build.sbt
+++ b/benchmarks/database/build.sbt
@@ -23,7 +23,7 @@ lazy val database = (project in file("."))
         ExclusionRule("org.slf4j", "slf4j-api")
       ),
       // Force newer JNA to support more platforms/architectures.
-      "net.java.dev.jna" % "jna-platform" % "5.9.0",
+      "net.java.dev.jna" % "jna-platform" % "5.10.0",
       // Add simple binding to silence SLF4J warnings.
       "org.slf4j" % "slf4j-simple" % "1.7.32"
     )
--- a/benchmarks/database/build.sbt
+++ b/benchmarks/database/build.sbt
@@ -21,7 +21,11 @@ lazy val database = (project in file("."))
         ExclusionRule("net.openhft", "chronicle-bytes"),
         ExclusionRule("net.openhft", "chronicle-threads"),
         ExclusionRule("org.slf4j", "slf4j-api")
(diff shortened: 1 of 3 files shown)


lbulej:
JNA [5.10.0](https://github.com/java-native-access/jna/releases/tag/5.10.0) with the fix to architecture mapping on SAP JVM (https://github.com/java-native-access/jna/commit/91eb8c27d21d646353a55e164d8788ecdb7242ce) was just released. I'm bumping the JNA version in this PR, which should address #153.

PR #327, opened 2021-11-09
Update apache-spark benchmarks to use Spark 3.2.0 and Scala 2.13.7

lbulej:
Benchmark code remains unchanged apart from relaxing imports for implicit conversions to Java collections and few other updates related to deprecations in Scala 2.13.


farquet (approved):

lbulej:
FYI, I'm waiting with the merge because I would like to update the Hadoop Windows binaries found at https://github.com/cdarlint/winutils. So far there is no `winutils.exe` and `hadoop.dll` for Hadoop 3.3 which Spark 3.2 is built against. The latest available binaries are for Hadoop 3.2.2 and the version we currently have in Renaissance is for Hadoop 3.2.0. I don't know how much of a problem it is.

farquet:
Regarding the Windows support, can't we use the builds linked in the issue here? -> https://github.com/cdarlint/winutils/issues/12

PR #336, opened 2021-12-09
Refactor build system 

lbulej:
This overhaul centralizes SBT configuration into a single `build.sbt`, which is more in line with how SBT was intended/recommended to be used for multi-project builds. Even though the main `build.sbt` is arguably large, a big chunk of it are project definitions which are now a bit shorter because we can now share some settings. Also, because I absolutely hate the implicit/hidden data flow between tasks which makes it difficult to figure out what is going on in such a complex system as SBT, I tried to write our custom tasks so that they are given the (most important) data explicitly in the project configuration.

Here's a summary of specific changes (those visible externally are marked in **bold**):

- **Benchmark modules (which group common dependencies) need to be added manually to the `renaissanceBenchmarks` variable.** I think it is a small cost, all things considered.
    - The way it was done before optimized for adding benchmarks quickly and (relatively) effortlessly, at the cost of twisting SBT into unintended usage modes, which make the SBT shell less useful, increase startup times and memory consumption. Apparently, this is not the operation we should optimize for at the expense of making the whole process impenetrable.
    - Because projects are now explicitly assigned to lazy vals, SBT can figure that out which makes switching between sub-projects in SBT shell actually usable/useful :-)
    - It is actually possible to query all defined projects and find out the benchmark ones, but it can be only done in a task the value of which can be only depended upon by other tasks. So this does not work for the `aggregate` thing we use on the main project. Without aggregation (which we probably don't need), it would be possible to query the defined benchmark projects and provide them to the tasks that need them. 

- We don't need to modify many files to update benchmark/build related settings.
    - Scala 2.12 and 2.13 point versions are shared by all benchmarks.
    - Plugins such as `sbt-scalafmt` only need to be configured once for the main project.
    - We can share dependency versions (e.g., JNA) between some of the projects.

- We can now build the `renaissance-harness` for Scala 2.12 and 2.13 and include both versions in the bundle.
    - The Launcher uses the 2.13 one because it will load it using a separate class loader, but the other version is meant to allow execution of individual benchmarks "outside" the bundle using a single class loader for everything, where we need the harness Scala version to match the version used by the benchmark.
    - More support for this will follow on top of this.

- Benchmarks are now actually compiled in parallel, not sequentially, which cuts down on compilation time a bit.

- The `sbt-assembly` plugin has been removed. It could not do what we wanted and we were just bending it sideways. It turns out that if we just provide the right mappings to the `package` task, it can do what we actually want.
  - When assembling the fat JAR bundles, we copy the files directly from the dependency cache instead of copying them first into the project target directory. This saves some IO.
  - We only put unique jar dependency files into the bundle, which reduces the bundle size by about 80 megs.
  - **There is no `assembly` task** &mdash; use the normal `package` task instead (or `renaissance/package` if you want to be specific).

- The `sbt-jmh` plugin has been removed too, because it was not helping much. I have replicated the steps it does using multiple projects (instead of twisting the configurations) so that it's now clear what is going on.
    -  It is somewhat easier to exclude the generator dependencies from the final bundle.
    -  We can update JMH independently of `sbt-jmh`.
    -  **There is no `assembly` task and no `Jmh` configuration** &mdash; use `renaissanceJmh/package` instead of `renaissanceJmh/Jmh/assembly` to build the JMH bundle.
    - **The name of the JMH bundle changed from `renaissance-jmh-assembly` to simply `renaissance-jmh`.**
        - Preserving the name requires an extra setting which I chose to avoid unless we really need it.

- **The support for running with debugger and the associated setting was removed.**
    -  I'm not sure it was even possible to run renaissance from SBT, because it has to be run from the bundle anyway. If somebody was using it, let me know.

- The whole process is also less sluggish (on my 4C/8T machine with nvme).
    - On a cold project (after git clean)
        - Startup in 27 instead of 39 seconds.
        - Base bundle compilation and packaging in 56 instead of 83 seconds.
        - JMH bundle compilation and packaging in 19 instead of 23 seconds.
    - On an initialized project
        - Startup in 12 instead of 19 seconds.
        - Base bundle compilation and packaging in 21 instead of 41 seconds.
        - Base bundle packaging (without compilation) in 16 instead of 22 seconds.
    - The CI has been updated to stop using the experimental SBT client/server (CI was failing with OpenJ9), and to make up for it, we build the base and JMH bundle in one SBT invocation (both bundles were built one after the other anyway).
    - SBT should use less memory (the code cache seems to fit to 128 megs again) and do less IO (even though it was copying files that were probably in page cache), so it might even shave some time from CI builds.

Regarding the "externally visible" changes, especially the change of task names &mdash; we could define a custom `assembly` task which would just execute the `package` task (command alias is not scoped), but I wanted to stick to essentials, given the complexity already imposed by SBT's build model. The same applies to custom configurations. If we need a stable build interface, we should define `renaissancePackage` and `renaissanceJmhPackage` command aliases to hide the details. 

I probably forgot something, but this should ease some of the SBT induced pain (it certainly does for me).


vhotspur (approved):
I am not a SBT expert but the changes look reasonable to me and the new `build.sbt` looks less daunting than before.

PR #341, opened 2021-12-20
Standalone mode support

lbulej:
Finally, here's an initial stab at support for running benchmarks outside the fat jar. Don't look at the huge commits, I'll split those up when we settle on the interface that suits you (and the dependency changes will go in separately). So here's how it works at the moment.

Build the bundle (as usual) but use the `--extract` option to extract a specific benchmark (you can extract all of them or a group -- the benchmark specifier works the same way as for running).

```shell
java -jar renaissance-gpl.jar --extract scala-kmeans
```

The harness will extract all the jars needed into the `scala-kmeans` subdirectory of the current directory. You can change the extraction base, i.e., instead of current directory, you can use `--extract-base /tmp`, which will create the `scala-kmeans` directory under `/tmp`. You cannot change the benchmark subdirectory name, but that should be fine.

The harness also creates a `standalone.jar` in the `scala-kmeans` directory, which only contains two files:
- The manifest from the original bundle (which means that the `Add-Opens` attribute is carried over) with some modifications. The `Main-Class` attribute contains the name of the harness class (instead of the launcher class), there is an additional `Class-Path` attribute which lists all the jar files needed to run the benchmark (including the core jar and the right harness jar), and there is a `Renaissance-Use-Modules` attribute (set to `false`) which tells the harness to simply load the benchmark class using the default class loader. 
- The `benchmarks.properties` file with metadata for the single benchmark.

The idea is to use the metadata-only jar to run the benchmark:

```shell
java -jar scala-kmeans/standalone.jar scala-kmeans
```

Or you can just use `all`, because the jar contains metadata only for the `scala-kmeans` benchmark anyway (running it with `--list` will only print details of the `scala-kmeans` benchmark).

If you don't want to use the `standalone.jar`, you can just assemble the class path from all the jar files in the `scala-kmeans` directory and run the main class `org.renaissance.harness.RenaissanceSuite`. You will also need to use the `--standalone` option to disable module loading and the `--benchmark-metadata` option to tell the harness where to find the benchmark metadata.
- The code currently does not do it, but if necessary, it could also create the `scala-kmeans/benchmarks.properties` file separately, so that everything that's needed to run without the standalone jar is there. However, in this mode, I don't see a way to pass the `Add-Opens` attribute from the main manifest. If you used `java -jar renaissance-harness_<version>.jar`, you cannot modify the class path, and if you run `java -cp <all-jars> mainClassName`, the `Add-Opens` attribute in the `renaissance-harness_*.jar` does not apply (I think). This was actually the reason for creating the metadata-only jar.

Can you (@farquet and @ceresek) try it out and see if the current interface (file/directory names, produced files, and options) fit your purposes?


lbulej:
In the meantime, I'll wrap the checked exceptions in the JMH wrapper :-)

ceresek:
@farquet This could replace the separate harness archive used by the mx AOT tooling, right ?

farquet:
Yes! That will make native image benchmarking a much smoother experience since the jar will provide everything needed.

lbulej:
BTW, options on how to go about this also include tweaking the build so that the standalone metadata-only jar is created for each module (or benchmark) during build and they would be put into a separate resource directory (say `standalone`) under unversioned name (e.g., `scala-kmeans.jar`) with suitable references to the `unique` and `shared` module directories.

Then you would just extract the main bundle, list the contents of the `standalone` directory to find a jar for the benchmark you want and run `java -jar` it.  The extraction and standalone jar creation code would be moved to the build system and we would get rid of the `--extract` and `--extract-base` options from the harness (we would only keep the `--standalone` and `--benchmark-metadata` flags for adventurers).

On the one hand, from a technical perspective, the build system might be the right place for this so as to avoid burdening the harness with unnecessary features. On the other hand, the current solution is kind of ready and does not involve SBT (even though I think this could be handled in reasonably obvious way using resource generators), but it really depends on what do you think would work best for you in the longer term.

For example, you cannot control the extraction subdirectory now, because it is possible to extract multiple benchmarks now. If that's not needed, the `--extract-base` option could become `--extract-dir` which would allow you to control exactly where you want to extract the benchmark, down to the last directory name. But then you would have to extract the benchmarks one by one (the specifier would have to identify a single benchmark) -- which might be just fine for use in scripts.

farquet:
First of all, awesome work! That will be highly valuable for a smooth experience benchmarking GraalVM native image.

I've played a bit with the `--extract` mechanism and it works very smoothly.

> BTW, options on how to go about this also include tweaking the build so that the standalone metadata-only jar is created for each module (or benchmark) during build and they would be put into a separate resource directory (say standalone) under unversioned name (e.g., scala-kmeans.jar) with suitable references to the unique and shared module directories.

> Then you would just extract the main bundle, list the contents of the standalone directory to find a jar for the benchmark you want and "java -jar" it. The extraction code would be drop, the standalone jar creation code would be moved to the build system and we would get rid of the --extract and --extract-base options from the harness (we would only keep the --standalone and --benchmark-metadata flags for adventurers).

I'm a big fan of this solution! The standalone jars are tiny because they only contain metadata, so they're cheap. And it would simplify the process.
IMO, we should go down that path because the current `--extract` mechanism - _while being many times better than the current design_ - is more complex and also opens a few questions (should we override if the directory exists? should we add a force mode? Should it be more verbose about what it does ? i.e. extract or skip).
Your suggestion of having one jar per benchmark in a `standalone` directory is the best option (at least for my use case). I'm not a big fan of the per-group solution since groups are either arbitrary (conceptually similar benchmarks) or technical (same project with the same set of dependencies). I think if one wants to iterate over a group (no need for me), it should use the `java -jar renaisssance.jar <group> --raw-list` and run the benchmarks listed. Interestingly this command currently returns the full list despite the group being given. Maybe there is another way to get the list of benchmarks from a group programmatically, but I couldn't find any.

Note that this new API would require a bit more testing in the context of native image to see if there are remaining glitches, but I don't see any reason to hold the merge.

lbulej:
> > BTW, options on how to go about this also include tweaking the build so that the standalone metadata-only jar is created for each module (or benchmark) during build and they would be put into a separate resource directory (say standalone) under unversioned name (e.g., scala-kmeans.jar) with suitable references to the unique and shared module directories.

> I'm a big fan of this solution! The standalone jars are tiny because they only contain metadata, so they're cheap. And it would simplify the process. IMO, we should go down that path because the current `--extract` mechanism - _while being many times better than the current design_ - is more complex and also opens a few questions (should we override if the directory exists? should we add a force mode? Should it be more verbose about what it does ? i.e. extract or skip). Your suggestion of having one jar per benchmark in a `standalone` directory is the best option (at least for my use case).

I'm also generally inclined towards this solution, even though it occurred to me only after I started creating the metadata-only jars to simplify the launch of the heap of extracted jars and wrote the initial PR description :-)

@ceresek Would that work for you as well?

ceresek:
Yo, either way is fine with me - in the infrastructure, we'll always have the full JAR and extract whatever is needed just before running the benchmark - and I think we'll be able to script around whatever details remain.

lbulej:
The bundle now contains a `single` directory with per-benchmark jars. After extracting the bundle, the metadata-only jars can be used to launch a single benchmark without module loading. The class path in the manifest is relative to the metadata-only jar location and points to the other directories with module dependency jars.

It seems to work as expected, so I removed the benchmark extraction and metadata-only jar generation code from the harness. What's left now are mostly finishing touches:

- [ ] Copy the main bundle manifest settings (especially `Add-Opens`, but the Git-relates ones too)
- [ ] Filtering the dummy benchmarks (except perhaps `dummy-empty`?)
- [ ] Any other tweaks you want

farquet (approved):

farquet:
Since this PR is close to be merged, I would suggest sneaking in the bump to Scala 2.13.8 which just got released.

https://github.com/scala/scala/releases/tag/v2.13.8

lbulej:
> Since this PR is close to be merged, I would suggest sneaking in the bump to Scala 2.13.8 which just got released and mainly contains bug fixes.

Sure thing.

lbulej:
> > Since this PR is close to be merged, I would suggest sneaking in the bump to Scala 2.13.8 which just got released and mainly contains bug fixes.
> 
> Sure thing.

On a second thought, I found one other easy update, so I'll move the version bumps into a separate PR on top of this. It will be easier to point out the PRs with the version changes :-)

lbulej:
OK, this seems to do what we planned, so I think we can move on for now. I'm keeping the original description so that the conversation that followed makes sense :-)

PR #343, opened 2021-12-20
Keep masking warnings from com.github.fommil.netlib

lbulej:
Francois @farquet noticed that even though Spark 3.2.0 uses `dev.ludovic.netlib`, the old logger still pops up, so we will set the log level in both for now. Let's see if we can clean this up next Spark update.


farquet (approved):
Thx

PR #346, opened 2022-01-27
Adding CI check for the standalone mode.

farquet:
It exhibits a problem with `dotty` which looks like a JDK compatibility issue:

```
====> dotty
```

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -50,6 +50,10 @@ jobs:
         shell: bash
         run: tools/ci/bench-base.sh
 
+      - name: Run the suite in standalone mode
+        shell: bash
+        run: tools/ci/bench-standalone.sh
+
       - name: Run the suite with JMH
         shell: bash
         run: tools/ci/bench-jmh.sh
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -39,6 +39,10 @@ jobs:
         shell: bash
         run: tools/ci/bench-base.sh
 
+      - name: Run the suite in standalone mode
+        shell: bash
+        run: tools/ci/bench-standalone.sh
+
       - name: Run the suite with JMH
         shell: bash
         run: tools/ci/bench-jmh.sh
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -70,6 +70,10 @@ jobs:
         shell: bash
         run: tools/ci/bench-base.sh
 
+      - name: Run the suite in standalone mode
+        shell: bash
+        run: tools/ci/bench-standalone.sh
+
       - name: Run the suite with JMH
         shell: bash
         run: tools/ci/bench-jmh.sh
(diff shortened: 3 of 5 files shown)


lbulej:
This is due to the fact that we build a class path for Dotty in the setup phase and while doing that, we assume that the class loader is the URLClassloader. It apparently isn't when the class path is defined in the manifest. Can you turn it into an issue?

Thanks for the CI check, LGTM.

farquet:
- [ ] Rebase the branch once this is fixed -> #347

vhotspur (approved):
Apart from the Dotty issue the patch looks good. Thanks!

farquet:
Rebasing on @lbulej 's PR, dotty isn't a problem anymore.

farquet:
I spoke too fast. It seems that now it's OpenJ9 on JDK8 who still has issues with this mechanism:
```
====> dotty
java.lang.NullPointerException
	at org.renaissance.scala.dotty.Dotty.classPathFromManifest$1(Dotty.scala:162)
	at org.renaissance.scala.dotty.Dotty.buildDottyClassPath(Dotty.scala:182)
	at org.renaissance.harness.RenaissanceSuite$$$Lambda$122/0x00000000f49f9350.apply(Unknown Source)
```

lbulej:
I noticed that too and tried to re-run the CI to see if it's fluke or not. I initially coded it with optionals to avoid null pointer exception when we fail to get a manifest input stream, but because that's something that should not really happen, I chose to let it fail if it happens so that we find out. But I did not expect to find out so quickly, because I don't see how that could fail: why, if you only have a single jar on the class path, could we not get manifest from the context class loader.

farquet:
Since this problem is only affecting a specific vendor on an old java version (JDK8) and in a specific mode (standalone). It's probably not critical to fix it (unless the fix is obvious, but it doesn't look like it).
So I would suggest either to drop `openj9-openjdk8` from the linux matrix, or to drop the standalone test on linux (it would also be tested in the main, macos and windows pipelines.

Any thought on this ?

I think we shouldn't hold the 0.14 release because of that.

lbulej:
> Since this problem is only affecting a specific vendor on an old java version (JDK8) and in a specific mode (standalone). It's probably not critical to fix it (unless the fix is obvious, but it doesn't look like it).

I have just downloaded OpenJ9 with JDK8, and it fails on my machine (but OpenJ9 with JDK11 does not) so I'll give it a quick look...

> So I would suggest either to drop openj9-openjdk8 from the linux matrix, or to drop the standalone test on linux (it would also be tested in the main, macos and windows pipelines.
> Any thought on this ?

It's what I would default to is no obvious fix.

> I think we shouldn't hold the 0.14 release because of that.

Agreed.

lbulej:
It turns out I was just stupid, see #349

ceresek:
Lucky we're talking past tense here :-)

lbulej:
> Lucky we're talking past tense here :-)

The past tense only applies to that particular problem... I reserve the right to be stupid in future :-)

lbulej:
One more rebase, for peace of mind.

PR #348, opened 2022-01-27
Fix standalone mode for Dotty

### LINKED ISSUE #347: Dotty fails in standalone mode.
```
====> dotty
```
@lbulej's [comment](https://github.com/renaissance-benchmarks/renaissance/pull/346#issuecomment-1023028891) on that:

> This is due to the fact that we build a class path for Dotty in the setup phase and while doing that, we assume that the class loader is the URLClassloader. It apparently isn't when the class path is defined in the manifest.

lbulej:
In standalone mode (without modules), even if the current class loader is a URLClassLoader (which it is on JDK8 but not on JDK11), it may not provide enough information about the real class path. For example, on JDK8, the class loader returns a single URL representing the jar file passed to 'java -jar', which is no different from the value of the 'java.class.path' system property.

So we instead try to detect if the benchmark is running in standalone mode and if yes, we build the class path using the 'Class-Path' attribute in the jar manifest. If this fails, we try to get class path URLs from the class loader, and in the worst case fall back to the current class path, which should help in situations where we try to run the benchmark completely manually, with all jar files listed on the command line.

Fixes #347

### DIFF
--- a/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
+++ b/benchmarks/scala-dotty/src/main/scala/org/renaissance/scala/dotty/Dotty.scala
@@ -17,13 +17,18 @@ import org.renaissance.core.DirUtils
 import java.io.File
 import java.io.FileInputStream
 import java.net.URLClassLoader
+import java.nio.file.Files
 import java.nio.file.Files.copy
 import java.nio.file.Files.createDirectories
 import java.nio.file.Files.notExists
 import java.nio.file.Path
+import java.nio.file.Paths
 import java.nio.file.StandardCopyOption.REPLACE_EXISTING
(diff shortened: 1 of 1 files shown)


lbulej:
Let's see if we can work around the issue like this. But if that's deemed too complicated, we can let benchmarks access the list of their dependency jars through benchmark context. However, because it is only needed by one benchmark at this time, I did not go that way.

farquet (approved):
Nice fix covering all the cases well.

Approving assuming you've also confirmed that it fixes the standalone mode problem.

lbulej:
> Approving assuming you've also confirmed that it fixes the standalone mode problem.

Yes, I tested specifically `dotty` both in bundle and standalone mode and on JDK8 and JDK11. Let's see if it can pass your CI check :-)

PR #349, opened 2022-01-28
Standalone mode: load manifest directly from jar, not as a resource

### LINKED ISSUE #347: Dotty fails in standalone mode.

lbulej:
Even though there is only a single jar on the class path, the class loader actually knows about all referenced jars and can search them for resources. Depending on the order in which the class loader searched the jars, we might get manifest from different jar instead of the one we are looking for. The fact that it was working so far was just luck. With newer JDKs, the primary jar (the only one on the explicit class path) was searched first, while on JDK8 it was not the case, so we were looking for the 'Class-Path' attribute in the wrong manifest.

Really fixes #347


farquet (approved):
Aaaah good catch!

PR #352, opened 2022-02-02
CI: test on JDK 18 and 19 too

vhotspur:
Using early-access tarballs (see [here](https://github.com/renaissance-benchmarks/docker-buildenv/blob/master/update.py#L87)).


lbulej (approved):
LGTM

PR #356, opened 2022-04-03
Avoid using iterator to access ArrayBuffer that can be modified during iteration

### LINKED ISSUE #354: ConcurrentModificationException on finagle-chirper
```
GC before operation: completed in 54.688 ms, heap usage 410.982 MB -> 34.636 MB.
[2022-02-21T16:17:56.179+0100] com.twitter.finagle.util.DefaultMonitor (com.twitter.finagle.util.DefaultMonitor logWithRemoteInfo)
WARNING: Exception propagated to the default monitor (upstream address: /127.0.0.1:56796, downstream address: n/a, label: ).

java.util.ConcurrentModificationException: mutation occurred during iteration
	at scala.collection.mutable.MutationTracker$.checkMutations(MutationTracker.scala:43)
	at scala.collection.mutable.CheckedIndexedSeqView$CheckedIterator.hasNext(CheckedIndexedSeqView.scala:47)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:926)
	at org.renaissance.twitter.finagle.FinagleChirper$Master.$anonfun$mostRechirpsInAllFeeds$1(FinagleChirper.scala:112)
	at org.renaissance.twitter.finagle.FinagleChirper$Master.$anonfun$mostRechirpsInAllFeeds$1$adapted(FinagleChirper.scala:110)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1293)
	at scala.collection.parallel.ParIterableLike$Foreach.leaf(ParIterableLike.scala:938)
	at scala.collection.parallel.Task.$anonfun$tryLeaf$1(Tasks.scala:52)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at scala.util.control.Breaks$$anon$1.catchBreak(Breaks.scala:97)
	at scala.collection.parallel.Task.tryLeaf(Tasks.scala:55)
	at scala.collection.parallel.Task.tryLeaf$(Tasks.scala:49)
	at scala.collection.parallel.ParIterableLike$Foreach.tryLeaf(ParIterableLike.scala:935)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.internal(Tasks.scala:169)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.internal$(Tasks.scala:156)
	at scala.collection.parallel.AdaptiveWorkStealingForkJoinTasks$AWSFJTWrappedTask.internal(Tasks.scala:304)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.compute(Tasks.scala:149)
	at scala.collection.parallel.AdaptiveWorkStealingTasks$AWSTWrappedTask.compute$(Tasks.scala:148)
	at scala.collection.parallel.AdaptiveWorkStealingForkJoinTasks$AWSFJTWrappedTask.compute(Tasks.scala:304)
	at java.base/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
```

I see this exception roughly every 20-30 iterations on darwin-aarch64 with graal-enterprise jarGraal. It doesn't seem to affect performance.

lbulej:
The switch to Scala 2.13 apparently unearthed an issue that has been present in `finagle-chirper` all the time. The code changes in `finagle-chirper` between Renaissance 0.13 and Renaissance 0.14 were only meant to "port" it to Scala 2.13 and required changing the signature of various methods that compute feed stats to accept a `SeqView` instead of `Seq` (because in Scala 2.13, a non-strict `SeqView` is no longer a subtype of a strict `Seq`), but even in Scala 2.12, the `view` method of an `ArrayBuffer` returned a `SeqView`.

The main cause appears to be the fact that the `analyze()` method accesses an `ArrayBuffer` (representing a per-user feed) without a lock. The feed is appended to under a lock in the `Master`, but no lock is held when calculating stats on individual feeds (in a `Future`). When processing a request for feed stats, the lock is held while creating a read only snapshot of the trie that holds all user feeds and a view is created for each feed (which essentially snapshots the number of items in the feed), but the view is processed later.

The `analyze()` function running as part of the future evaluation uses an iterator to go over chirps in a feed, but the underlying ArrayBuffer can be modified (appended to) in the meantime. In Scala 2.13 this results in the `ConcurrentModificationException`. I tested the same code in Scala 2.12 (with the same libraries) and I was not able to reproduce the problem.

Now, I believe that the race is intentional and should be benign (similar to updating profiling counters in hotspot), so to keep it the way it was, I just changed the `analyze()` method to avoid using an iterator and instead iterate over a `Range` and then access the `ArrayBuffer` elements by index. This should be "safe" even if the `ArrayBuffer` is expanded (and the backing array changed) because the `ArrayBuffer` is only appended to. Using a direct index should either fetch the element from an old array or the new array, but the returned item should be the same &mdash; assuming the implementation updates the reference to the backing array **after** it has copied the contents of the old array to the new array.

* Strictly speaking, I'm not sure whether even this would hold under the memory model, but I understand the desire to avoid locking for these computations &mdash; the whole point of the stats is to provide some reasonable numbers at some point in time and if some updates come in between, they will be included in the next stats call.

Fixes #354 (as long as keeping the race can be considered a fix)

### DIFF
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
@@ -53,17 +53,22 @@ final class FinagleChirper extends Benchmark {
     var requestCount = 0
     var postCount = 0
 
-    def analyze[T](feed: SeqView[String], zero: T, f: String => T, op: (T, T) => T): T = {
+    def analyze[T](
+      feed: IndexedSeqView[String],
+      zero: T,
+      f: String => T,
+      op: (T, T) => T
+    ): T = {
       val a = new Accumulator(zero)(op)
       for (msg <- feed) a.accumulate(f(msg))
       a.get()
     }
 
-    def longestMessageInFeed(feed: SeqView[String]): String = {
+    def longestMessageInFeed(feed: IndexedSeqView[String]): String = {
       analyze[String](feed, "", x => x, (x, y) => if (x.length > y.length) x else y)
     }
 
-    def longestMessageInAllFeeds(allFeeds: Seq[SeqView[String]]): String = {
+    def longestMessageInAllFeeds(allFeeds: Seq[IndexedSeqView[String]]): String = {
       var result = ""
       for (feed <- allFeeds) {
         val r = longestMessageInFeed(feed)
@@ -87,7 +92,7 @@ final class FinagleChirper extends Benchmark {
       result
     }
 
-    def longestRechirpInAllFeeds(allFeeds: Seq[SeqView[String]]): String = {
+    def longestRechirpInAllFeeds(allFeeds: Seq[IndexedSeqView[String]]): String = {
       var result = ""
       for (feed <- allFeeds) {
         val s = analyze[String](
@@ -105,7 +110,7 @@ final class FinagleChirper extends Benchmark {
       result
     }
 
-    def mostRechirpsInAllFeeds(allFeeds: Seq[(String, SeqView[String])]): Long = {
+    def mostRechirpsInAllFeeds(allFeeds: Seq[(String, IndexedSeqView[String])]): Long = {
       val counts = ConcurrentHashMultiset.create[String]()
       allFeeds.par.foreach {
         case (username, feed) =>
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
@@ -118,7 +118,7 @@ final class FinagleChirper extends Benchmark {
       counts
         .entrySet()
         .parallelStream()
-        .map[Integer]((t: Entry[String]) => t.getCount)
+        .map[Integer] { t: Entry[String] => t.getCount }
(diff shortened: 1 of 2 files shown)


farquet (approved):
Very nice analysis!
I agree that it makes sense not to add locking for a stats feature. Your fix is probably how one would want to solve this in an actual production workload.
I've tested your branch locally and numbers are comparable with renaissance 0.14 as expected.

Issue #345, opened 2022-01-26
finagle-chirper reports exceptions during execution on Zing VM


ceresek:
While executing (like, once every 10ish repetitions), finagle-chirper reports exceptions when running on Zing VM:

```
====== finagle-chirper (web) [default], iteration 43 started ======                                                                                                                       [53/1878]
Resetting master, feed map size: 5000                                                                                                                                                              
GC before operation: completed in 94.977 ms, heap usage 3.402 GB -> 592.000 MB.                                                                                                                    
[2022-01-26T09:52:51.475+0000] com.twitter.finagle.util.DefaultMonitor (com.twitter.finagle.util.DefaultMonitor logWithRemoteInfo)                                                                 
WARNING: Exception propagated to the default monitor (upstream address: /127.0.0.1:48902, downstream address: n/a, label: ).                                                                       
                                                                                                 
java.util.ConcurrentModificationException: mutation occurred during iteration
        at scala.collection.mutable.MutationTracker$.checkMutations(MutationTracker.scala:43)
        at scala.collection.mutable.CheckedIndexedSeqView$CheckedIterator.hasNext(CheckedIndexedSeqView.scala:47)
        at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
        at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
        at scala.collection.AbstractIterable.foreach(Iterable.scala:926)       
        at org.renaissance.twitter.finagle.FinagleChirper$Master.analyze(FinagleChirper.scala:58) 
        at org.renaissance.twitter.finagle.FinagleChirper$Master.$anonfun$hashStartCountInAllFeeds$1(FinagleChirper.scala:84)
        at org.renaissance.twitter.finagle.FinagleChirper$Master.$anonfun$hashStartCountInAllFeeds$1$adapted(FinagleChirper.scala:77)
        at scala.collection.immutable.List.foreach(List.scala:333) 
        at org.renaissance.twitter.finagle.FinagleChirper$Master.hashStartCountInAllFeeds(FinagleChirper.scala:77)
        at org.renaissance.twitter.finagle.FinagleChirper$Master.$anonfun$apply$9(FinagleChirper.scala:184)
        at com.twitter.util.Try$.apply(Try.scala:26)             
        at com.twitter.util.ExecutorServiceFuturePool$$anon$4.run(FuturePool.scala:163)
[...]

PR #357, opened 2022-04-09
Iterate over a snapshot of the ArrayBuffer internals

lbulej:
This is a follow up to #356 and provides an alternative fix to #354 (and #345) that should be compatible with the Java memory model. Here we use a `FeedBuffer` class derived from `mutable.ArrayBuffer` to store the feed strings. The derived class provides a `snapshotView()` method that creates a snapshot view of the underlying array (instead of the buffer). This ensures that even if the buffer is appended to (and the underlying array is replaced), we can still iterate over the original contents. This works under the assumption that the feed represented by the buffer is only appended to.

The `snapshotView()` method is called under the same lock that is used to modify the buffer contents and we don't need to worry about the underlying array reference being changed during the iteration. Let me know if you think I overlooked something.

### DIFF
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
@@ -31,6 +31,8 @@ import java.util.Comparator
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 import scala.collection._
+import scala.collection.immutable.ArraySeq
+import scala.collection.mutable.ArrayBuffer
 import scala.collection.parallel.CollectionConverters._
 import scala.io.Source
 import scala.util.hashing.byteswap32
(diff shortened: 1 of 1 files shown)


farquet:
This looks like a reasonable and safe approach. The consequence is the extra object allocation every time `snapshotView()` is called.
Have you checked its impact on memory pressure and on peak performance ? That's especially interesting in a CI-like setup where Xmx is low.
Those are short-lived objects which will be collected quickly, so it's probably ok?

lbulej:
> The consequence is the extra object allocation every time snapshotView() is called.

Yes, there is an extra `ArraySeq` being allocated to wrap the old array reference. We could take a snapshot of just the reference and the current number of items and iterate over the array directly, but it would feel much less like Scala code.

> Have you checked its impact on memory pressure and on peak performance ? That's especially interesting in a CI-like setup where Xmx is low.

I did not check with specific Xmx settings, but the amounts of data freed between repetitions are very similar between 0.14, the previous fix and this fix.

I did a cursory check (i.e., not many runs) on the (default 90) repetition times, and the average for this fix seems to be on par with the 0.14 release.

farquet (approved):

PR #361, opened 2022-05-25
GHA: various small improvements

vhotspur:
## Base cache id on main sbt only 

No need to recursively scan all directories.

Furthermore, it seems to cause spurious errors with some runners where some of the temp directories were inaccessible.

Actually, using cache id of `{{ os }}-sbt` should be enough as the versions are handled by SBT anyway and there is no need to drop the whole cache when upgrading one package but this is probably safer.

## Continue on errors

Since GitHub allows (finally) to restart a single job, it seems better to allow all jobs to continue (even if one of them fails) as many of the failures are spurious ones anyway.

## Use `-` instead of `/dev/stdout`

Somehow the use of `/dev/stdout` caused hard-to-explain errors on Windows runners (when executing `tools/ci/bench-show-env.sh`). Now the `ResultWriter` explicitly handles `-` to print to standard output (as is usual for many Unix utilities). As an extra benefit, we do not close stdout this way.

## Workflow chaining

Run the `Main` workflow first and run the other ones (more JDKs etc.) only if it succeeds.

This probably would need to be merged first (according to the [second note here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)).

## Reducing CI jobs

The `Main` workflow was reduced to run on `master` branch and PRs only. Therefore, we should not see that for an opened PR there are jobs related to the branch _and_ to the PR that are actually testing exactly the same code.

One can always create a draft PR to check the branch (otherwise, mere `git push` will not trigger a CI).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -34,7 +34,7 @@ jobs:
             ~/.ivy2/cache
             ~/.sbt
             ~/.cache/coursier
-          key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
+          key: ${{ runner.os }}-sbt-${{ hashFiles('build.sbt') }}
 
       - name: Environment configuration
         shell: bash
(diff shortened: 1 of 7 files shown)


lbulej (approved):
LGTM

Issue #331, opened 2021-11-12
Streamline CI on GHA


vhotspur:
Follow-up to the discussion in #325 ([direct link](https://github.com/renaissance-benchmarks/renaissance/pull/325#pullrequestreview-803676465)).

Main point is to test LTS versions first.

farquet:
For reference, the early access builds from Oracle can be found on dockerhub at `openjdk:<jdk>-ea-oracle`.

Example: `docker pull openjdk:19-ea-oracle`

lbulej:
Also, is there a way to build the main workflow first, so that it does not have to wait for the others to finish?

vhotspur:
As a follow up on #362.

At the moment, we build everything (all the old JDK versions and plugins) on each push (or PR).

Practically, the jobs need about 10-15 minutes each, all of them are done in about an hour. The main four jobs (general checks for formatting, latest JDK on Linux, Windows and MacOS) are usually done in 25 minutes.

If we want to cut down the time, we need to decide what to run and when.

There are several options available where to split things (i.e. when to decide which jobs to execute). Interestingly, Linux job takes about 10 minutes, while Windows or Mac need about 13 minutes (and the scheduling for these seems to be generally slower).

* Run all checks on `master` only (we merge only after PR, hence there is less of time pressure and it speeds-up getting results in PRs).
* Run all checks on tagged versions only (i.e. test only releases on on all JDKs). Generally this would speed-up things but it may require bisecting in the end.
* Run all checks when commit message contains something like `[ci:all]`.
* Run all checks periodically (e.g. every week).
* Run all checks on PRs only when review was requested.
* Built JAR only once and share it between the jobs (the build alone takes about 3 minutes). This is something I have not investigated in much detail but it should be possible. We would probably build with JDK8 and test building on other JDKs optionally only?

We can also combine the above options somehow or drop the tests for old JDKs completely (IMHO we should keep them for testing releases at least).

@lbulej What makes the most sense for you as you probably use it the most?

@farquet, @axel22 From your perspective, what is the bare minimum where we want to test always (for example, would latest JDK on Linux only be sufficient)?

farquet:
We have some benchmarks that are JDK11 only, so testing only the latest JDK would ignore those.
Moreover, the fact that the JDK changes can be significant and can impact the functionality of the suite, it feels safer to test at least LTS releases we support.

[...]

PR #362, opened 2022-05-26
[draft] GHA: another attempt for dependencies (see #331)

vhotspur:


lbulej:
Seems to be going the right way ;-) What will be used for the status icon on the Renaissance website? When the files were split, we used the "main" workflow, which reported only the result of building on LTS 17 on Linux/MacOS/Windows. From what will be derived the result of this workflow?

BTW, given how slow the builds are, consider building only on LTS version (8, 11, 17) + plus the most recent one (i.e. 18)?

vhotspur:
@lbulej Can we merge this? I made in mistake in the last PR (#361) as the dependent workflows are actually always run on `master` and never actually touch the code of the branch in the PR.

Right now, it reports 20 minutes duration which is quite fine and it fails fast.

lbulej (approved):
> @lbulej Do you think we can merge this? I made in mistake in the last PR (#361) as the dependent workflows are actually always run on `master` and never actually touch the code of the branch in the PR. Thanks!

LGTM. Just keep an eye on the website button.

> >  BTW, given how slow the builds are, consider building only on LTS version (8, 11, 17) + plus the most recent one (i.e. 18)?
> Right now, it reports 20 minutes duration and it fails fast. Do we want to drop the other versions completely?

My point is, I'm not sure we would care too much if it failed anywhere between LTS versions and was fine with the LTS versions and the latest versions. So maybe it will not be faster, but we will waste less resources. It might be an interesting thing to do for releases (i.e. tags), but I don't find it of much value for normal development. Part of it may be that the (former) linux workflow with other JDKs has been failing so much recently (despite my attempts at restarting the builds), that I started to completely ignore it.

vhotspur:
> > @lbulej Do you think we can merge this? I made in mistake in the last PR (#361) as the dependent workflows are actually always run on `master` and never actually touch the code of the branch in the PR. Thanks!
> 
> LGTM. Just keep an eye on the website button.

Will do.


> > > BTW, given how slow the builds are, consider building only on LTS version (8, 11, 17) + plus the most recent one (i.e. 18)?
> > > Right now, it reports 20 minutes duration and it fails fast. Do we want to drop the other versions completely?
> 
> My point is, I'm not sure we would care too much if it failed anywhere between LTS versions and was fine with the LTS versions and the latest versions. So maybe it will not be faster, but we will waste less resources. It might be an interesting thing to do for releases (i.e. tags), but I don't find it of much value for normal development. Part of it may be that the (former) linux workflow with other JDKs has been failing so much recently (despite my attempts at restarting the builds), that I started to completely ignore it.

That is a good point. I will merge this as it fixes the current state and will have a look at having the other jobs available for tags only.

PR #366, opened 2022-09-22
Bump slf4j to safe version

### LINKED ISSUE #365: Renaissance uses old slf4j version that ships a vulnerable log4j version
The currently used slf4j version 1.7.33 uses a vulnerable log4j version, see https://www.slf4j.org/news.html.
This is problematic when running renaissance on servers in our intranet (and leads many security scanners to flag our renaissance instance as problematic).

parttimenerd:
Bump slf4j to version that does not ship a vulnerable log4j version (see https://www.slf4j.org/news.html). Version 1.7.36 ships with reload4j instead of the affected log4j.

This essentially prevents security scanners from flagging the renaissance benchmark suite as a security risk.

Fixes  #365

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -135,7 +135,7 @@ val jnaVersion = "5.10.0"
 val nettyVersion = "4.1.72.Final"
 val scalaCollectionCompatVersion = "2.6.0"
 val scalaParallelCollectionsVersion = "1.0.4"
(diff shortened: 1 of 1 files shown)


vhotspur (approved):

vhotspur:
@parttimenerd Merged, thanks!

PR #367, opened 2022-10-05
Remove log4j dependencies

lbulej:
Attempt to completely remove `log4j` from `apache-spark` dependencies. Complements #365 which updated `slf4j-log4j12` to pull in `reload4j`, but did not remove direct the Spark dependency on `log4j`.  With `log4j` removed, there seem to be some rough edges (missing appenders), probably configuration related. If this can be sorted out, we can do a minimal-change maintenance release, otherwise we'll see if the problems go away on its own after updating most of the dependencies.

### DIFF
--- a/benchmarks/actors-akka/src/main/resources/reference.conf
+++ b/benchmarks/actors-akka/src/main/resources/reference.conf
@@ -1,0 +1,15 @@
+# Override configuration in actors-akka.jar
+
+akka {
+
+  # Log level used by the configured loggers (see "loggers") as soon
+  # as they have been started; before that, see "stdout-loglevel"
+  # Options: OFF, ERROR, WARNING, INFO, DEBUG
+  loglevel = "WARNING"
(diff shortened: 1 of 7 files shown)


lbulej:
I have moved Spark's log4j configuration file to a place where `reload4j` can find it, so we should be back to normal logging configuration.

lbulej:
With a few minor updates, the suite seems to be free of `log4j` and does not print logging noise.

Overall, logging seems to be a somewhat messy issue, especially with libraries using different logging solutions. We should be able to get somewhat "unified" control over dependencies that use `org.slf4j`, but there are still some that use `java.util.logging` directly (which cannot be configured by just placing a properties file on the class path).  I'll revisit this after updating major libraries, but first I would like to get a log4j-free maintenance release (with updated JNA) out.

farquet (approved):
LGTM

Issue #369, opened 2022-11-15
Can we sync with the jna community to support the LoongArch architecture?


Panxuefeng-loongson:
Hi all,

I am a developer from Loongson. LoongArch is a risc architecture developed by Loongson. Renaissance is an authoritative benchmark, we use it to improve the quality of the Loongson platform JDK, but the jna package supporting LoongArch is missing in renaissance, here is the contents of jna unpacked in renaissance:
```
loongson@loongson-pc:~/下载/shared/temp/com/sun/jna$ ls linux-
linux-aarch64/  linux-arm/      linux-armel/    linux-mips64el/ linux-ppc/    
linux-ppc64le/  linux-riscv64/  linux-s390x/    linux-x86/      linux-x86-64/
```
Now that LoongArch is supported by the jna community: https://github.com/java-native-access/jna/pull/1440, can we sync it so that LoongArch users can also use renaissance easily?

Thanks

Panxuefeng-loongson:
I am glad to receive your reply. Bumping JNA to version 5.12.1 will solve the problem. I'm looking forward to supporting LoongArch's renaissance.

Panxuefeng-loongson:
> > I am glad to receive your reply. Bumping JNA to version 5.12.1 will solve the problem. I'm looking forward to supporting LoongArch's renaissance.
> 
> Renaissance now uses JNA 5.12.1 (see release 0.14.2), so I hope that it fixes the problem.

Good news to me, thanks

lbulej:
I'm assuming the update fixed the problem. Please reopen if necessary.

PR #371, opened 2023-02-08
Update JNA to version 5.12.1

lbulej:
This should be a minimal compatibility change to take care of #369. We should be able to make a compatibility release before moving on to more substantial library/framework updates.


vhotspur (approved):
LGTM

Issue #375, opened 2023-03-04
Document possible StackOverflowError on the Spark benchmarks


farquet:
The context is the following OpenJDK bug that was been closed as a non-issue: https://bugs.openjdk.org/browse/JDK-8303076

Under certain circumstances (special JVM configurations like the one described in the bug), it is possible that Spark benchmarks fail with a _StackOverflowError_. The reason is that the benchmarks themselves allocate a lot on the stack and are already close to the overflow limit. This seems to be a common issue with Spark applications.

The simple workaround is to increase the stack size with `-Xss`. That would be great if this workaround would be documented somewhere in the repo (main README?).

Hopefully opening this issue already helps potential users hitting this in the future.

```
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
NOTE: 'als' benchmark uses Spark local executor with 12 (out of 12) threads.
====== als (apache-spark) [default], iteration 0 started ======
GC before operation: completed in 33.258 ms, heap usage 179.102 MB -> 38.409 MB.
23/02/22 14:54:31 WARN GarbageCollectionMetrics: To enable non-built-in garbage collector(s) List(G1 Concurrent GC), users should configure it(them) to spark.eventLog.gcMetrics.youngGenerationGarbageCollectors or spark.eventLog.gcMetrics.oldGenerationGarbageCollectors
23/02/22 14:54:50 ERROR Executor: Exception in task 8.0 in stage 30.0 (TID 247)
java.lang.StackOverflowError
at org.apache.spark.util.ByteBufferInputStream.read(ByteBufferInputStream.scala:49)
at java.base/java.io.ObjectInputStream$PeekInputStream.read(ObjectInputStream.java:2911)
at java.base/java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2927)
at java.base/java.io.ObjectInputStream$BlockDataInputStream.readInt(ObjectInputStream.java:3448)
at java.base/java.io.ObjectInputStream.readHandle(ObjectInputStream.java:1866)
at java.base/java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:1927)
at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2248)
at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1760)
at java.base/java.io.ObjectInputStream$FieldValues.<init>(ObjectInputStream.java:2614)
[...]

PR #377, opened 2023-03-23
CI: use versioned image names

vhotspur:
So far, images from the `docker.io/renaissancebench/buildenv` were versioned by JDK version only. That makes testing of new images impractical as there is no way to distinguish releases of the images (e.g. the image name does not reflect which Fedora version was used).

Thus pushing new version to DockerHub immediately influences how we build _any_ version of Renaissance here on GitHub CI.

The `v1-` version is a tagged version of existing images and hence the transition should be smooth.

`v2-` would come in a separate PR and would probably introduce images built on Fedora 37 and for newer JDKs too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   checks:
     runs-on: ubuntu-latest
-    container: "renaissancebench/buildenv:openjdk8"
+    container: "renaissancebench/buildenv:v1-openjdk8"
     steps:
       - name: Git checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


vhotspur:
@lbulej The Spark benchmarks are continuing to fail on JDK19 because of StackOverflow. Seems like #375 instance. Have you encountered this before, please?

Otherwise it looks healthy and ready for review.

farquet (approved):
LGTM

lbulej:
> The Spark benchmarks are continuing to fail on JDK19 because of StackOverflow. Seems like https://github.com/renaissance-benchmarks/renaissance/issues/375 instance. Have you encountered this before, please?

Some time ago when dealing with Spark benchmarks. I was also playing with "StackSmash" test to see if perhaps we would warn about it automatically, but nothing conclusive came out of it. In general, it fits the behavior described in #375.

Issue #384, opened 2023-05-05
Dotty exception starting in JDK21-b21


ericcaspole:
Starting in JDK-21-b21, due to this fix https://bugs.openjdk.org/browse/JDK-8292275 Dotty fails to run in 0.14.1. and 0.14.2.

I don't know much about Scala or Dotty, thanks for your analysis.

```
jdks/21-b21/Contents/Home/bin/java -jar renaissance-jmh-0.14.2.jar Dotty -t 1 -i 15 -wi 75 -f 1
# JMH version: 1.34
# VM version: JDK 21-ea, Java HotSpot(TM) 64-Bit Server VM, 21-ea+21-LTS-1778
# VM invoker: /Users/ecaspole/dev/opt/brunte/jdks/21-b21/Contents/Home/bin/java
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 75 iterations, single-shot each
# Measurement: 15 iterations, single-shot each
# Timeout: 10 min per iteration
# Threads: 1 thread
# Benchmark mode: Single shot invocation time
# Benchmark: org.renaissance.scala.dotty.JmhDotty.runOperation

# Run progress: 0.00% complete, ETA 00:00:00
# Fork: 1 of 1
# Warmup Iteration   1: dotty-error: error while loading ElementType,
class file /modules/java.base/java/lang/annotation/ElementType.class is broken, reading aborted with class java.lang.RuntimeException
bad constant pool index: 0 at pos: 1220
<failure>

org.renaissance.BenchmarkResult$ValidationException: compilation errors: expected 0 but got 1
	at org.renaissance.BenchmarkResult$Assert.assertEquals(BenchmarkResult.java:167)
	at org.renaissance.scala.dotty.Dotty.$anonfun$run$1(Dotty.scala:228)
	at org.renaissance.jmh.JmhRenaissanceBenchmark.tearDownOperation(JmhRenaissanceBenchmark.java:135)
	at org.renaissance.scala.dotty.jmh_generated.JmhDotty_runOperation_jmhTest.runOperation_SingleShotTime(JmhDotty_runOperation_jmhTest.java:510)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
[...]

Issue #370, opened 2023-01-27
NoSuchMethodException: java.nio.DirectByteBuffer.<init>(long,int) with newest JDK 21-b7


ericcaspole:
We are using 0.14.1.
Starting from JDK 21-b7, this exception happens related to the bug fix for  https://bugs.openjdk.org/browse/JDK-8299684

If this is fixed in Renaissance, possibly by a Spark upgrade, in newer unreleased code please let me know and we will plan for that.

java.lang.ExceptionInInitializerError
at org.apache.spark.unsafe.array.ByteArrayMethods.<clinit>(ByteArrayMethods.java:54)
at org.apache.spark.internal.config.package$.<clinit>(package.scala:1149)
at org.apache.spark.SparkConf$.<clinit>(SparkConf.scala:654)
at org.apache.spark.SparkConf.set(SparkConf.scala:94)
at org.apache.spark.SparkConf.set(SparkConf.scala:83)
at org.apache.spark.sql.SparkSession$Builder.$anonfun$getOrCreate$1(SparkSession.scala:920)
at scala.collection.mutable.HashMap$Node.foreach(HashMap.scala:627)
at scala.collection.mutable.HashMap.foreach(HashMap.scala:489)
at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:920)
at org.renaissance.apache.spark.SparkUtil.setUpSparkContext(SparkUtil.scala:121)
at org.renaissance.apache.spark.SparkUtil.setUpSparkContext$(SparkUtil.scala:92)
at org.renaissance.apache.spark.Als.setUpSparkContext(Als.scala:43)
at org.renaissance.apache.spark.Als.setUpBeforeAll(Als.scala:88)
at org.renaissance.jmh.JmhRenaissanceBenchmark.setUpBenchmark(JmhRenaissanceBenchmark.java:115)
at org.renaissance.apache.spark.jmh_generated.JmhAls_runOperation_jmhTest._jmh_tryInit_f_jmhals0_G(JmhAls_runOperation_jmhTest.java:590)
at org.renaissance.apache.spark.jmh_generated.JmhAls_runOperation_jmhTest.runOperation_SingleShotTime(JmhAls_runOperation_jmhTest.java:477)
at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
at java.base/java.lang.reflect.Method.invoke(Method.java:578)
[...]

PR #386, opened 2023-05-17
Temporarily restrict supported JVM version in benchmarks broken changes in JDK21

lbulej:
Changes in JDK21 break apache-spark benchmarks and the scala-dotty benchmark (see #370 and #384). With the apache-spark benchmarks, we are waiting for a fix that is expected to appear in Spark 3.5. With the scala-dotty benchmark, we need to check if the issue is handled upstream. For the time being (to allow including JDK21 in the CI testing matrix), we will restrict the supported JVM in these benchmark to JDK20.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,35 +67,35 @@ The following is the complete list of benchmarks, separated into groups.
 
 - `als` - Runs the ALS algorithm from the Spark ML library.
   \
-  Default repetitions: 30; APACHE2 license, MIT distribution; Supported JVM: 1.8 and later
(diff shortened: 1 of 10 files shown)


lbulej:
> Note that both Dotty and Spark addressed the problem upstream. So as soon as they do proper releases (namely Spark 3.5.0 and Dotty 3.3.1), we could bump those versions and release a renaissance 0.15.0 with JDK21 compatibility.

Yes. The intention of the fix is to allow testing JDK21-ea in CI. When we update Dotty and Spark to the upstream versions, the restrictions will be removed.

Issue #372, opened 2023-02-24
Suspected memory leak in bench Reactors


dhanalla:
Encountered the following OOM with java -Xms4G -Xmx4G  -jar renaissance-jmh-0.14.1-10.jar -bm avgt -prof gc -f 1 -t 20 -i 20 Reactors

java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.concurrent.ForkJoinTask.adapt(ForkJoinTask.java:1499)
        at io.reactors.JvmScheduler$ReactorForkJoinWorkerThread.executeLater(JvmScheduler.scala:120)
        at io.reactors.JvmScheduler$ReactorForkJoinWorkerThread.postschedule(JvmScheduler.scala:162)
        at io.reactors.JvmScheduler$Executed.postschedule(JvmScheduler.scala:273)
        at io.reactors.concurrent.Frame.executeBatch(Frame.scala:202)
        at io.reactors.JvmScheduler$Executed$$anon$5.run(JvmScheduler.scala:255)
        at java.base/java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1375)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)

memory is held for more than 110 generations. This issue reproduces on Windows & Linux .
![image](https://user-images.githubusercontent.com/45609141/221225256-8a618ac3-0089-4604-b244-3fb805a8f875.png)

lbulej:
Thanks for the report. Can you provide more info on the JDK you were running the benchmark on?
@ceresek did you observe something like that in the long runs?

ceresek:
Our long runs typically used 12G heap (Xms, Xmx) and G1, and never encountered an OOM (with OpenJDK 11 most recently), so this configuration must be different.

ceresek:
Most likely fixed in 0.15.0, please reopen if it happens again ?

PR #389, opened 2023-08-09
Plugging reactor leaks

ceresek:
The leaks were in the `ForkJoinThroughput` and `Fibonacci` workloads, the former retained around a megabyte of heap data per repetition, the other was somewhat smaller.

### DIFF
--- a/benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
+++ b/benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
@@ -273,6 +273,7 @@ object Fibonacci {
     system.spawn(Reactor[Int] { self =>
       self.main.events.onEvent { x =>
         done.success(x)
+        self.main.seal()
       }
       fib(self.main.channel, sz)
     })
@@ -367,7 +368,7 @@ private object ForkJoinThroughput {
         })
       }).toArray
 
-    system.spawn(Reactor[String] { _ =>
+    system.spawn(Reactor[String] { self =>
       var j = 0
       while (j < sz) {
         var i = 0
@@ -377,6 +378,7 @@ private object ForkJoinThroughput {
         }
         j += 1
       }
+      self.main.seal()
     })
 
     var res: Int = 0


benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
```diff
@@ -273,6 +273,7 @@ object Fibonacci {
     system.spawn(Reactor[Int] { self =>
       self.main.events.onEvent { x =>
         done.success(x)
+        self.main.seal()
```
axel22, line 276: Good catch.

vhotspur (approved):

benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
lbulej: Good catch indeed!

Just one thing &mdash; in other benchmarks we seal the channel before completing the future. I don't know if that would be a problem here, it just caught my eye.

lbulej:
This should help with/fix #372, right?

benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
ceresek: Feel free to regularize as you see fit :-) Alex is the best authority on this, my feeling is sealing of the main channel will only have an effect after the event handler returns so it should not matter ...

lbulej: Your ordering might even be the better one, if I consider channel as a resource that should be cleaned up after the primary activity (future completion) of the handler. I have a stash of various cleanups in `reactors` from April/May, when I was looking for the race, so I'll do something about it when I get to making a PR out of it :-)

Feel free to merge.

lbulej (approved):

benchmarks/actors-reactors/src/main/scala/org/renaissance/actors/Reactors.scala
axel22: I think that the order inside `onEvent` should not matter much (but if you put it before, you will potentially very slightly decrease the number of events that are pushed onto that channel's queue, because it's closed sooner).

axel22: (IIRC as soon as `seal` is called, all subsequent events delivered to that channel are dropped)

ceresek: I think all the workloads only send events when the recipient has not yet closed their channel (the communication pattern is such that everybody either only reacts or knows when to stop), so there should really be no difference.

PR #394, opened 2023-09-12
Streamline CI

vhotspur:
Three major changes:

* CI runs only on LTS and latest JDK
  * Not JDK21-ea as we need to upgrade SBT and perhaps something else too?
* Switch to `v5` of the Docker images (new rebuilds, added OpenJ9 for 17, new Ant etc.)
* Make build on latest EA JDK on-demand only (not fully tested, these jobs can be executed only on master)

Overall, this should reduce CI time significantly while still keeping a wide range of tested JDK versions.

### DIFF
--- a/.github/workflows/ea-jdk.yml
+++ b/.github/workflows/ea-jdk.yml
@@ -1,0 +1,53 @@
+name: EA JDK
+on:
+  push:
+  schedule:
+    - cron: "17 23 17 * *"
+
+jobs:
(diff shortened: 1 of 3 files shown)


lbulej:
LGTM

lbulej:
Regarding JDK21, I tried with SBT 1.9.4 (latest at the moment), but so far without success.

lbulej (approved):

PR #398, opened 2023-09-20
Update build configuration, SBT and JMH

lbulej:
This is the first in a series of PRs towards next release. This contains some tweaks to the build configuration so that we can propagate common options more easily and to differentiate between options for different Scala versions. SBT is updated so that we stay current and can eventually build on JDK21 (a few more updates may be needed though).


vhotspur (approved):
LGTM.

PR #400, opened 2023-09-20
Update Neo4j to version 4.4.25

lbulej:
In addition to the update, this also extracts JSON loading to a higher level (from `AnalyticsBenchmark` to `Neo4jAnalytics`) and uses `play-json` instead of `lift-json` to load JSON data. The data itself has been sanitized with respect to types and the structure has been simplified, dropping unnecessary field identifiers to simplify loading.

Also includes a bunch of cleanups to avoid Scala style violations and/or some compiler warnings.

### DIFF
--- a/benchmarks/neo4j/src/main/scala/org/renaissance/neo4j/Neo4jAnalytics.scala
+++ b/benchmarks/neo4j/src/main/scala/org/renaissance/neo4j/Neo4jAnalytics.scala
@@ -72,7 +72,7 @@ final class Neo4jAnalytics extends Benchmark {
   }
 
   override def tearDownAfterAll(c: BenchmarkContext): Unit = {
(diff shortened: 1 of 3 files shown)


vhotspur (approved):
Looks okay. I really like the shift towards explicit model (classes).

farquet:
That's unfortunate to be stuck at JDK20 max version for the neo4j benchmark while JDK21 is now out.

Is it a lot of effort to migrate from Neo4J 4 to more recent version? If that's the case, we should try to get some help from Neo4J team. Being JDK21+ compatible really allows frameworks to be thoroughly tested and optimized by JVM and compiler developers.

cc @jexp

lbulej:
I'll check if we can migrate to Neo4J 5.x to get JDK21 compatibility at the cost of requiring JDK17.

PR #403, opened 2023-09-20
Introduce Scala 3 variant of the harness

lbulej:
Requires a few tweaks to allow cross-compiling between Scala 2.12, 2.13 and 3.3. This is needed to enable gradual migration of (some) benchmarks to Scala 3. This is needed by a subsequent update to `dotty` to avoid incompatibility with JDK21.


vhotspur (approved):
I am glad I do not have to write this myself :-)

PR #405, opened 2023-09-20
Migrate scala-stdlib to Scala 3

lbulej:


vhotspur (approved):
I just hope that all the change would also work together ;-)

lbulej:
> I just hope that all the change would also work together ;-)

They should :-)

PR #409, opened 2023-09-21
Update shared dependencies to use more recent versions

lbulej:
These dependencies are shared by multiple (groups of) benchmarks. We use the same versions across the whole suite to reduce the size of the final bundle.


vhotspur (approved):
As long as it compiles :-)

ceresek:
... and passes validation, right ?

Issue #408, opened 2023-09-21
neo4j-analytics fails with JDK21


lbulej:
The benchmark was previously limited to JDK15. After update to Neo4j 4.4.25, I was able to lift the restriction (PR coming), but not completely&mdash;the benchmark fails with JDK21 with the following chain of exceptions (see [neo4j.log](https://github.com/renaissance-benchmarks/renaissance/files/12682220/neo4j.log)):

```
java.lang.RuntimeException: Error starting Neo4j database server at [...]
Caused by: org.neo4j.kernel.lifecycle.LifecycleException: Component 'org.neo4j.dbms.database.DefaultSystemGraphInitializer@6afbe6a1' was successfully initialized, but failed to start. Please see the attached cause exception "null".
Caused by: org.neo4j.graphdb.DatabaseShutdownException: This database is shutdown.
Caused by: java.lang.RuntimeException: Failed to wrap pointer in ByteBuffer.
Caused by: java.lang.UnsupportedOperationException
	at java.base/java.lang.invoke.VarForm.getMemberName(VarForm.java:118)
	at org.neo4j.internal.unsafe.UnsafeUtil.initDirectByteBuffer(UnsafeUtil.java:665)
	at org.neo4j.internal.unsafe.UnsafeUtil.newDirectByteBuffer(UnsafeUtil.java:648)
	at org.neo4j.io.pagecache.impl.SingleFilePageSwapper.convertToByteBuffers(SingleFilePageSwapper.java:518)
```

The same code works just fine with JDK20. For now, I'll limit supported JVM in `neoj4-analytics` to version 20 (better than 15), but version 21 is a no-go. I did not try with Neo4j 5.x, because the benchmark code needs to be updated to use it.

lbulej:
Fixed by moving to Neo4j 5.12 (#422)

PR #410, opened 2023-09-21
Bump supported JVM in neo4j-analytics to version 20

lbulej:
This expands the range of JVM versions supported by `neo4j-analytics` to [11,20] instead of [11,15], but JDK21 is a no-go for now (see #408).


vhotspur (approved):
:+1:

PR #412, opened 2023-09-30
Update primary dependencies in db-shootout benchmark to improve JVM compatibility

lbulej:
Updates MapDB version from 3.0.1 to 3.0.10, H2 MVStore from version 1.4.192 to version 2.1.214, and Chronicle Map from version 3.17.0 to version 3.22.9

This makes the `db-shootout` benchmark compatible with JDK 18 (up from 11). 

Starting with JDK 19, the benchmark fails because `chronicle-core` is using reflection to look up the `sun.nio.ch.FileChannelImpl.map0(int,long,long,boolean)` method which is not available in JDK 19 anymore.

Using more recent Chronicle Map (such as 3.23.5 or even 3.24ea3) is currently not an option, because SBT fails to find their `third-party-bom`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ The following is the complete list of benchmarks, separated into groups.
 
 - `db-shootout` - Executes a shootout test using several in-memory databases.
   \
-  Default repetitions: 16; APACHE2 license, MIT distribution; Supported JVM: 1.8 - 11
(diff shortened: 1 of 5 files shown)


vhotspur (approved):
Tested on *OpenJDK 64-Bit Server VM (build 18.0.1.1+2-6, mixed mode, sharing)*, looks good. Thanks!

PR #413, opened 2023-10-03
Use Scala 3 version of the harness by default and in Java-only benchmarks

lbulej:
Because we need the harness to build with Scala 3 anyway, this makes Scala 3 the default for the harness and for the Scala wrappers of Java-only benchmarks. Unlike the wrappers, the harness needs to be maintained to cross-build with older Scala versions (2.13 and 2.12) so that an appropriate harness version can be used with the standalone benchmarks.


vhotspur (approved):
LGTM.

PR #414, opened 2023-10-04
Update RxJava to version 3.1.8 and port rx-scrabble to rxjava3

lbulej:
The benchmark code includes a number of small cleanups prior to porting it to the `rxjava3` API. Most changes are simple renames, and there are few changes which take advantage of methods available in `rxjava3`. Many streams now use the `Single` or `Maybe` type, which better fits their purpose and also avoids various conversions, because the `collect()` and `reduce()` operations in `rxjava3` produce those types by default. The benchmark appears to run faster compared to the version using version 1 API.


farquet:
new API looks indeed much nicer and readable. LGTM.

> The benchmark appears to run faster compared to the version using version 1 API.

What's the scale of the speedup? Assuming it's not order of magnitudes faster, I assume there is no need to adjust the workload size ?

farquet (approved):

lbulej:
> > The benchmark appears to run faster compared to the version using version 1 API.
> 
> What's the scale of the speedup? Assuming it's not order of magnitudes faster, I assume there is no need to adjust the workload size ?

Just to give you some ballpark figures, the average of the last 10 repetitions out of 80 gives me:

| Variant | JDK8 [ms] | JDK11 [ms] | JDK17 [ms] | JDK21 [ms] |
|:--------|------------:|-------------:|-------------:|-------------:|
| rxjava1 with `Observable` only (the original version from Renaissance 0.14.2) | 234 | 271 |  255 | 263 |
| rxjava1 with `Single`, `Observable`, with some unnecessary wrapping/mapping removed | 263 | 314 | 327 | 321 |
| rxjava3 with `Single`, `Maybe`, `Observable`, with operations such as `flattenAsObservable` | 109 | 132 | 143 | 125 |

So there is a binary order of magnitude. In general, increasing the workload size would not hurt, but it's not critical.

lbulej:
I have updated the table to include rxjava3 numbers with `Observable` only and with the extra wrapping, and it does represent some overhead, but it does not seem to be the cause of the speedup.

lbulej:
OK, so unless we want to keep the overhead due to extra wrapping (maybe it was intentional in the original version?), I would go ahead with this. Perhaps @axel22 knows more about the original intent?

lbulej:
I'll merge it as is for now, if Alex comes up with a reason to keep using `Observable` only and the extra wrapping of iterables, I can always add it later (I have the code for that locally).

PR #416, opened 2023-10-05
Update README.md

BlankSpacePlus:
While reading the documentation information, I found some typos or inconsistent formatting. So I tried to fix them.


vhotspur (changes requested):
Hello, thanks for the pull request (I have already enabled CI for it).

Unfortunately, we cannot merge it in its current state as some files are generated and one needs to modify the sources too (e.g. `README.md` is generated and you need to build Renaissance to generate the updated version).

Thank you.

vhotspur (approved):
Thanks a lot!

PR #417, opened 2023-10-06
Minimize build.sbt for the included reactors dependency

lbulej:
Drops cross-building and tests. Avoids import failures on conflicting dependencies when opening the project with IntelliJ IDEA. Might speed up SBT initialization by dropping cross-build plugins that produce additional projects.


lbulej:
> Really nice to reduce the amount of code so much. It looks reasonable now, I wonder how we managed to have so complex code there in the first place :-D

That was the original `build.sbt` from the reactors project which was only slightly modified to pick up common `scalaVersion` from the parent project. To point of including it (almost) as is was to have relatively minimal difference between the original project and the dependency we include, but once IDEA started to complain about import, I realized we can probably live with a different/simplified build config :-)

I wonder whether we should also delete the sources for the subprojects we don't use...

PR #418, opened 2023-10-06
Update neo4j to version 4.4.26

lbulej:
We might as well use the latest version at the moment...


lbulej:
I realized I have a few other trivial dependency updates, so I'll bundle them together with this one another PR.

PR #422, opened 2023-10-11
Update Neo4j to version 5.12.0 and migrate to Scala 2.13

lbulej:
This makes `neo4j-analytics` compatible with JDK21, but requires JDK17 to run.


farquet (approved):
Awesome!

PR #423, opened 2023-10-12
Use common versions of various dependencies

lbulej:
Also drops a few unused transitive dependencies. Reduces the size of the bundle.


lbulej:
I hope this to be the last PR before release :-)

vhotspur (approved):

PR #425, opened 2023-10-20
Update Hadoop to version 3.3.6

lbulej:
This is to improve compatibility with (more recent) IBM Semeru builds of OpenJ9-based JVMs. There was some renaming going on that resulted in changes to the contents of `java.vendor`, resulting in Hadoop 3.3.4 failing to find some classes. This was fixed in Hadoop version 3.3.5, but 3.3.6 is the most recent.

Details at https://github.com/apache/hadoop/pull/4537


lbulej:
Merging to move on.

PR #431, opened 2023-12-03
Add validation to fj-kmeans benchmark

lovisek:
For each loop, computed centroids are sorted by the value of their first dimension. After sorting, every centroid position is compared with expected position of given centroid. If expected centroid positions equals to the computed centroid positions, we can validate the benchmark result.

### DIFF
--- a/benchmarks/jdk-concurrent/src/main/scala/org/renaissance/jdk/concurrent/FjKmeans.scala
+++ b/benchmarks/jdk-concurrent/src/main/scala/org/renaissance/jdk/concurrent/FjKmeans.scala
@@ -4,8 +4,11 @@ import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
-import org.renaissance.BenchmarkResult.Validators
+import org.renaissance.BenchmarkResult.Assert
+import org.renaissance.BenchmarkResult.ValidationException
 import org.renaissance.License
+
+import java.{lang => jl, util => ju}
 
 @Name("fj-kmeans")
 @Group("jdk-concurrent")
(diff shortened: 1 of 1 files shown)


benchmarks/jdk-concurrent/src/main/scala/org/renaissance/jdk/concurrent/FjKmeans.scala
```diff
@@ -4,9 +4,12 @@ import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
-import org.renaissance.BenchmarkResult.Validators
+import org.renaissance.BenchmarkResult.Assert
+import org.renaissance.BenchmarkResult.ValidationException
 import org.renaissance.License
 
+import java.{lang => jl, util => ju}
```
farquet, line 11: I don't think this shortening style is used anywhere else in the project.
For code readability, I would have a preference to keep explicit `java.util` and `java.lang` in code. Especially since we're not saving that many characters here.

farquet (commented):
Looks good to me. Thanks for the contribution!

Could you quantify the performance impact (if any) and report here?

farquet (approved):
LGTM! Thanks!

lovisek:
I quantified performance impact by running benchmark before and after validation and computing average duration time from last 5 benchmark iterations.
average iteration duration time without validation: 3837,074ms
average iteration duration time with validation: 3717,384ms
Looks like validation doesn't have impact on benchmark performance.
[fj-kmeans.no-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/13661413/fj-kmeans.no-validation.result.txt)
[fj-kmeans.with-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/13661418/fj-kmeans.with-validation.result.txt)

PR #432, opened 2023-12-04
Add validation to future-genetic benchmark

lovisek:
There were race conditions in the computation of future-genetic, because all threads in algorithm parallelization were using same random generator. Therefore results were not stable. Solved by creating and adding to the computation engine new instances of Alterer, Survivor Selector, Offspring selector, Genotype factory and initialized those instances with their own random generator. (Created new instances of original objects, but with unique random generator connected to it.)

Validation is done by sorting genes of result genotype by their value and comparing genes values with expected values of genes in the genotype.

### DIFF
--- a/benchmarks/jdk-concurrent/src/main/java/org/renaissance/jdk/concurrent/JavaJenetics.java
+++ b/benchmarks/jdk-concurrent/src/main/java/org/renaissance/jdk/concurrent/JavaJenetics.java
@@ -1,23 +1,25 @@
 package org.renaissance.jdk.concurrent;
 
-import java.util.Random;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.IntStream;
-
+import io.jenetics.Alterer;
 import io.jenetics.Chromosome;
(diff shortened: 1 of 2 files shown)


farquet:
Thanks a lot @lovisek for the contribution! The PR looks good to me.

Before merging, could you confirm that this has no impact on performance? Before and after your change, do the iteration times look roughly similar?

farquet (approved):

lovisek:
I quantified performance impact by running benchmark before and after validation and computing average duration time from last 5 benchmark iterations.
average iteration duration time without validation: 1568,813ms
average iteration duration time with validation: 1091,311ms
Looks like after changes is benchmark even faster.
[future-genetic.no-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/13661525/future-genetic.no-validation.result.txt)
[future-genetic.with-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/13661531/future-genetic.with-validation.result.txt)

lbulej:
> average iteration duration time without validation: 1568,813ms average iteration duration time with validation: 1091,311ms Looks like after changes is benchmark even faster.

That's interesting. Even the very first (cold) iteration is faster.  One significant change this can be attributed to is that the worker threads no longer contend for a single instance of `Random`.

PR #435, opened 2024-02-10
Add validation to als benchmark

lovisek:
The trained ALS model can be represented as a 2-dimensional matrix of features and the validation checks the shape and the content of the matrix. Regarding the content, it compares the sum and sum of squares of all matrix values with expected values.

The validation does not change the code that is being measured and should not really impact benchmark performance. As a quick check, I calculated an average duration from the last 5 repetitions of a single run of the benchmark before and after validation:

- without validation: 2235,345ms
([als.no-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/14229213/als.no-validation.result.txt))
- with validation: 2323,872ms
([als.with-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/14229214/als.with-validation.result.txt))

The variant with validation appears approx. 4% slower, but this is just from a single run. Given that the benchmark code did not change, I don't think this indicates an actual change, but if necessary, I can do the comparison for multiple runs.

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/Als.scala
@@ -8,7 +8,7 @@ import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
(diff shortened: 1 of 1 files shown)


farquet (approved):
LGTM! Thanks @lovisek for the contribution.

PR #436, opened 2024-02-10
Add validation to log-regression benchmark

lovisek:
Validation currently supports only binary classification which returns a single intercept value. If multinomial logistic regression is needed, the validation needs to be updated to support multiple intercept values.

The validation checks logistic regression model's coefficients, intercept, and the number of classes the model can classify. For the coefficients, it compares the sum and sum of squares of all coefficient values with the expected values, and then validates the coefficient count. Regarding the intercept, its value is compared with the expected value, and then the validation checks that the model contains only one intercept.

The validation does not change the code that is being measured and should not really impact benchmark performance. As a quick check, I calculated an average duration from the last 5 repetitions of a single run of the benchmark before and after validation:

- without validation: 920,128ms
[log-regression.no-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/14229679/log-regression.no-validation.result.txt)
- with validation: 929,781ms
[log-regression.with-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/14229677/log-regression.with-validation.result.txt)

The variant with validation appears approx. 1% slower, but this is just from a single run. Given that the benchmark code did not change, I don't think this indicates an actual change, but if necessary, I can do the comparison for multiple runs.

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/LogRegression.scala
@@ -7,11 +7,10 @@ import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
-import org.renaissance.BenchmarkResult.Validators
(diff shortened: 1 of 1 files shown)


farquet (approved):
LGTM!
Thanks for the contribution.

PR #437, opened 2024-02-26
add cpu usage plugin resolves #238

### LINKED ISSUE #238: How to use JMX plugin
Hi,

I see that as part of the new release, it is possible to gather and report JMX metrics within the benchmark.

This can be very handy, however, I couldn't find any example about how to use them. I tried using the various command line options with no luck.

Could you kindly provide a working example?

Thank you!

BakhtinArtem:
Based on the request in post #238, I added a plugin that exports CPU metrics. Unfortunately, I ran into a problem with inconsistent measurements. The `OperatingSystemMXBean` module,  which is used to gather the metrics, can return `NaN` values if the CPU load cannot be read at that moment.  This produces a `0` values  in the measurement results. 

Also, not all devices provide the `OperatingSystemMXBean` subinterface of the `com.sun.management` package, so some metrics are not available in some OSes.

### DIFF
--- a/plugins/jmx-cpu/README.md
+++ b/plugins/jmx-cpu/README.md
@@ -1,0 +1,59 @@
+# jmx-cpu plugin for Renaissance suite
+
+This plugin collects information about processor usage 
+via [OperatingSystemMXBean](https://docs.oracle.com/javase/8/docs/api/java/lang/management/OperatingSystemMXBean.html).
+
+## Building
+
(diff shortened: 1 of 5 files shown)


lbulej:
I'll merge this into a separate branch for now, I want to check if it's better to report negative numbers or `null` in the JSON instead of 0 to represent the `NaN` that could be returned by the `getSystemLoadAverage()` method.

PR #438, opened 2024-03-12
docs: remove shell prefix in CONTRIBUTING.md

guilhas07:
Remove the shell prefix so that the user can just click the copy icon and paste the command.


vhotspur (approved):
Makes sense, thanks for the patch!

I never noticed the icon until now :-D.

PR #441, opened 2024-04-23
Add validation to movie-lens benchmark

lovisek:
The benchmark originally returned an empty set of recommended movies due to passing a userID of value 0 to the predict method. User ID 0 is undefined (as the model was not trained with ratings from this user), causing the predict method to return an empty set. This issue was resolved by retrieving the userID from the 'ratings-personal.csv' file.

The validation process involves two steps. First, it checks whether the specified number of top recommended movies contains a defined set of expected movies. Next, the validation compares the Root Mean Square Error (RMSE) achieved by the best model on the validation subset with the expected value.

The validation does not change the code that is being measured to a greater extent and thus should not really impact benchmark performance. As a quick check, I calculated an average duration from the last 5 repetitions of a single run of the benchmark before and after validation:

without validation: 6916.339ms
([movie-lens.no-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/15082502/movie-lens.no-validation.result.txt))
with validation: 6973.107ms
([movie-lens.with-validation.result.txt](https://github.com/renaissance-benchmarks/renaissance/files/15082514/movie-lens.with-validation.result.txt))

The variant with validation appears approx. 1% slower, but this is just from a single run. If necessary, I can do the comparison for multiple runs.

### DIFF
--- a/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala
+++ b/benchmarks/apache-spark/src/main/scala/org/renaissance/apache/spark/MovieLens.scala
@@ -9,7 +9,8 @@ import org.renaissance.Benchmark
 import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
-import org.renaissance.BenchmarkResult.Validators
+import org.renaissance.BenchmarkResult.Assert
+import org.renaissance.BenchmarkResult.ValidationException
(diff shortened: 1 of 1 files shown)


lbulej:
LGTM, but I was involved prior to the PR, so I would welcome someone else to take a peek :-)

farquet (approved):
Looks good to me!

Thanks @lovisek for the contribution.

PR #442, opened 2024-04-28
Chat application benchmark using Kotlin with Ktor framework (Closes #428)

### LINKED ISSUE #428: Renaissance meets Kotlin
## Introduction

Renaissance is quite a diverse and popular benchmarking suite for JVM, which covers Scala and Java with popular libraries and use cases. However, it lacks benchmarks for a trendy JVM language - Kotlin. This work aims to solve this issue by creating and integrating the Kotlin benchmark in the form of a simple server application using Ktor.

## Why Ktor?

The benchmark goal is to utilize as much of the code in the specific language (to be precise, JVM bytecode compiled from a specific language) in the popular use case. Having that in mind, here are the reasons for the usage of Ktor:

1. Most frameworks for server-client applications like Spring or Retrofit are written in Java and only provide some shortcut wrappers in Kotlin. If we write an application using such a framework, the resulting bytecode mostly comes from Java, which is not what we want. Ktor, on the other hand, is written in Kotlin, so all of the resulting bytecode would be coming from the Kotlin compiler.
2. Having either a server or a client that would send/receive HTTP requests is common throughout all kinds of user and business applications - such as web, mobile, and desktop apps or some cloud services. At the same time, Kotlin is becoming increasingly popular for Multiplatform application development (where server/client code is utilized a lot), with companies like Netflix using it. That is why a server application is perfect for the benchmark.

zavidnyi:


README.md
```diff
@@ -25,7 +25,7 @@ To run a Renaissance benchmark, you need to have a JRE version 11 (or later)
 installed and execute the following `java` command:
 
 ```
-$ java -jar 'renaissance-gpl-0.15.0.jar' <benchmarks>
+$ java -jar 'renaissance-gpl-0.16.0.jar' <benchmarks>
```
farquet, line 28: @lbulej, should the version bump happen in such a PR? I think we usually do not touch the version number for features/benchmarks, but only do when we actually do a release?

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+The benchmark is designed to simulate a chat application where multiple clients send requests simultaneously.
+It uses the [Ktor](https://ktor.io/) framework.
+
+Namely, the benchmark runs multiple clients at the same time.
+Clients are set up with one or both of the following tasks:
+
+- Join a chat and send a randomized message to it.
+- Create a private chat with another user (if not already created) and send a randomized message to them.
+
+They repeat those task the required number of times.
```
farquet, line 12: Could you clarify what is the required number of times? Is it `iterations_count`?

Typo: tasks

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+After that number of successfully completed tasks is validated.
```
farquet, line 13: This sentence is unclear. Please rephrase.

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+It's also possible to implement validation via end-to-end comparison of the artifacts is produced.
```
farquet, line 15: I'm not sure to understand this sentence either. What are those artifacts?

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Running the benchmark
+
+To run benchmark, execute the following command:
+
+```bash
+java -jar <renaissance.jar> kotlin-ktor
+```
+
+Here are the arguments that can be passed to the benchmark:
```
farquet, line 30: I would say "parameters" or "options" instead of "arguments". Arguments usually implies positional arguments on the CLI.

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+- `client_count`: Number of clients that are simultaneously sending the requests.
+  Increasing number of clients (with proportional decrease in number of repetitions) doesn't have much effect on the
+  overall runtime. Default value is the number of available CPUs.
+- `iterations_count`: Number of times clients should repeat their designated operations. Default value is 2000.
```
farquet, line 35: One operation is always one message, right? If that's the case, maybe we should name this parameter `messages_count` to avoid any confusion with benchmark iterations.

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+The high-level idea is that you of coroutines is that multiple blocking operations may run on the same thread pool.
+For more details, please see [the official coroutines guide](https://kotlinlang.org/docs/coroutines-overview.html)
+This is exactly what Ktor uses.
+
+Unfortunately, the Ktor framework does not provide a way neither to limit the number of threads used for internal
+operations nor to have separate thread pools for the client and server operations.
+Ktor, mostly uses the `Dispatchers.IO`and `Dispatchers.Default`, which operate on intersecting thread pools.
+The `Dispatchers.Default` has a thread pool of size equal to number of CPU cores, but it's at least 2.
+The `Dispatchers.IO` is a bit more complicated.
+By default, it has a thead pool of size 64 (which could be controlled via a system
```
farquet, line 57: Typo: thread

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+property`kotlinx.coroutines.io.parallelism`).
```
farquet, line 58: Missing space

benchmarks/kotlin-ktor/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+It's possible to influence the amount of contention present by adjusting the parameters of the bencmark.
+Increasing the number of chats will decrease contention for the modification of the chat history (the opposite is true
+as well).
+Increasing group/private message fraction will increase contention for the chat history and the chat pool, respectively.
+Increasing the number of clients will increase contention in all synchronization points.
+
+## Core classes overview
+
+- `KtorRenaissanceServer` is the root class which initialises, runs and tears down the server and the clients.
```
farquet, line 85: Isn't `KtorRenaissanceBenchmark`?

benchmarks/kotlin-ktor/src/main/kotlin/org/renaissance/kotlin/ktor/KtorRenaissanceBenchmark.kt
```diff
@@ -0,0 +1,120 @@
...
+import org.renaissance.kotlin.ktor.server.ChatApplication
+import kotlin.math.min
+
+@Group("web")
+@Name("kotlin-ktor")
+@Summary("Simple Ktor chat application with multiple clients, performing various tasks.")
+@Licenses(License.MIT)
+@Parameter(
+  name = "port",
+  defaultValue = "8080",
```
farquet, line 21: Could we use a less common port by default?

benchmarks/kotlin-ktor/src/main/kotlin/org/renaissance/kotlin/ktor/KtorRenaissanceBenchmark.kt
```diff
@@ -0,0 +1,120 @@
...
+    clientCount = context.parameter("client_count").toPositiveInteger()
+    numberOfRepetitions = context.parameter("iterations_count").toPositiveInteger()
+    val numberOfChats = context.parameter("chat_count").toPositiveInteger()
+    fractionOfClientsSendingGroupMessages = context.parameter("group_message_fraction").toDouble()
+    fractionOfClientsSendingPrivateMessages =
+      context.parameter("private_message_fraction").toDouble()
+    val randomSeed = context.parameter("random_seed").toPositiveInteger()
+
+    application = ChatApplication(numberOfChats)
+    server = embeddedServer(io.ktor.server.cio.CIO, host = "127.0.0.1", port = port) {
```
farquet, line 78: You should use `localhost` instead of `127.0.0.1`, otherwise it may fail on IPv6-only environments.

benchmarks/kotlin-ktor/src/main/kotlin/org/renaissance/kotlin/ktor/client/Client.kt
```diff
@@ -0,0 +1,76 @@
...
+  private val tasksToRun: List<ClientTask>,
+) {
+  /**
+   * Runs the client by establishing a WebSocket connection and performing the specified tasks.
+   *
+   * @return The number of successful tasks. Should be equal to `operationsRepetitions * tasksToRun.length`
+   */
+  suspend fun run(): Int {
+    var successfulTasks = 0
+    httpClient.webSocket(method = HttpMethod.Get, host = "127.0.0.1", port = port, path = "/ws/${userId}") {
```
farquet, line 34: Same comment as above regarding the use of `127.0.0.1`

benchmarks/kotlin-ktor/src/main/kotlin/org/renaissance/kotlin/ktor/client/ClientManager.kt
```diff
@@ -0,0 +1,57 @@
...
+import io.ktor.client.*
+import io.ktor.util.*
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlin.random.Random
+
+/**
+ * A **probabilistic** client manager, which set's up the appropriate amount of clients, performing a specified number and kind of tasks
```
farquet, line 12: Typo: sets

farquet (commented):
Thanks a lot @zavidnyi for the contribution!

The benchmark looks very much in line with the spirit of the suite and the implementation matches well the rest of the suite.

README.md
lbulej: It should not. I think this might be the result of not pulling tags from the remote when cloning the repository. In such case, I think the version from `version.sbt` will be used, which is always one step ahead of the current version. Ilia, if you  pull the tags to your local clone, rebuild renaissance and then run the archive with `--readme` command, it should generate an updated `README.md` (with the benchmark) but without the version change.

benchmarks/kotlin-ktor/README.md
zavidnyi: No, one operation could be represented by multiple messages.
For example, to send a private message to another user, you may first need to create a private chat with them.
To create a chat, a `CreateDirectMessageChatCommand` must be sent first.

I'll rephrase from "operations" to "tasks"

zavidnyi:
Hey @farquet,
I tried to clarify the benchmark readme.
Let me, know if it's more clear now or if there any other issues

zavidnyi:
@lbulej I've fixed an issue by disabling the Koltin plugin for both `renaissanceJmhWrappers` and dependant `renaissanceJmh.`
However, this seems like more of a workaround rather than a fix.
I'm unfamiliar with sbt, so it's hard for me to say why `koltinCompile` even triggers there.

lbulej:
> @lbulej I've fixed an issue by disabling the Koltin plugin for both `renaissanceJmhWrappers` and dependant `renaissanceJmh.` 

Thanks!

> However, this seems like more of a workaround rather than a fix. I'm unfamiliar with sbt, so it's hard for me to say why `koltinCompile` even triggers there.

No problem, that's the usual case with SBT :-) I'll check later if there is a better way to do this, but I want to merge the benchmark in a state that passes the CI.

One (I believe) last thing: please add `test` and `jmh` configurations to the benchmark &mdash; see for example [RxScrabble.scala](https://github.com/renaissance-benchmarks/renaissance/blob/master/benchmarks/rx/src/main/scala/org/renaissance/rx/RxScrabble.scala). The `test` configuration is supposed to trigger a short workload meant to be run during CI. The `jmh` configuration is used when run via JMH and is usually empty (meaning it uses the defaults) &mdash; unless there is a need for an override under JMH.

zavidnyi:
@lbulej I've added the `test` and `jmh` configurations for the `kotlin-ktor`

One question:
Where is exactly `jmh` configuration is used? 
I've not found any place in the repository which calls it and according to the `README` I'd assume that I can pass arguments to the JMH package as usual, without extra empty configuration

lbulej:
> @lbulej I've added the `test` and `jmh` configurations for the `kotlin-ktor`

Thanks!

> One question: Where is exactly `jmh` configuration is used? I've not found any place in the repository which calls it and according to the `README` I'd assume that I can pass arguments to the JMH package as usual, without extra empty configuration

It is used (by default) by the [JmhRenaissanceBenchmark](https://github.com/renaissance-benchmarks/renaissance/blob/65d596ea92d9e24e624b94dab6391c5f0801fb0e/renaissance-jmh/wrappers/src/main/java/org/renaissance/jmh/JmhRenaissanceBenchmark.java#L51) class, which is the base class for JMH wrapper classes that are generated for individual benchmarks.

lbulej:
> Hi @lbulej, I've just noticed that PR still not merged. I can see that some checks failed, but for some reason I can't see the details. Could you please rerun them, so I could check what's wrong?

The JVM has trouble finding some Kotlin classes when run under JMH. I have re-run the checks so that you could see for yourself (you should be able to run the CI checks in your repository too).

nomisRev:
@lbulej could you run the checks again? The logs have disappeared, I was trying to investigate to fix the build and get this merged ☺️ I discussed with @zavidnyi, and he currently doesn't have any time to invest in this effort.

PR #444, opened 2024-05-26
Switch to external scala-stm and update to version 0.11.1

lbulej:
Switching to external `scala-stm` allows eliminating the `scala-stm` sources from the repository, retaining just the adaptor code for `stmbench7`, which is not updated upstream. We now also compile the slightly modified `stmbench7` (Java) sources during build, instead of using the precompiled jar. This also opens the path for moving the `scala-stm` benchmarks to Scala 3, which will come in subsequent PR.


farquet (approved):
reviewed commit by commit.
Looks good to me!

PR #446, opened 2024-05-27
Add simple validation to the philosophers benchmark

lbulej:
This started as a simple hash-based validation @BohdanQQ's [fork](https://github.com/BohdanQQ/renaissance/tree/benchmark/philosophers/validation) from which I cherry-picked bits. Because there is no reason to validate hash of the output when we have the actual values, I modified the validation to validate the number of meals eaten directly and added an extra check that the forks have no owner at the end.


lbulej:
Rebased to master after merging #444.

farquet (approved):
LGTM. Thanks!

axel22 (approved):

Issue #202, opened 2019-09-17
Philosophers - clusters in iteration execution time


axel22:
Here are some warmup curves related to the `philosophers` benchmark:

![screenshot-17-51-16](https://user-images.githubusercontent.com/151774/65058342-670f4b00-d974-11e9-9e6c-9a6a413ed15f.png)

We should potentially investigate reasons for this - which might be that the `CameraThread` is doing a different number of iterations depending on the timing.

axel22:
cc @ceresek

Theodor:
I wonder what amount of threads (==philosophers) are these graphs for?

ceresek:
This was measured on Xeon E5-2620 v4 with hyperthreading turned off, so the JVM saw 8 processors, which is the default philosopher count for the benchmark.

PR #447, opened 2024-06-01
Make the number of "camera scans" constant in the philosophers benchmark

lbulej:
This includes two related changes and one extra bit.

The first reduces the duration of the transaction building the "image of the scene". Previously the image was being built inside the transaction, making the transaction last unnecessarily longer. Now we only take a snapshot of the state within the transaction and build the image outside the transaction to avoid blocking philosopher threads.

The second change makes the number of scans constant for a given workload. Previously, camera scans were triggered in reference to wall clock time, resulting in variable number of scans during each repetition. Now the camera scans are requested based on thread progress. To keep the number of camera scans constant, the requests are counted to avoid potential loss of wake-ups when the rate of progress increases (due to reduced contention towards the end of the workload).

These two changes should alleviate the banding in the operation durations (#202), but probably not entirely, because there simply is a thread that periodically blocks all threads for a while.

The extra bit is inclusion of the number of camera scans in the validation.

Below are plots from simple measurements I collected from 15 runs (150 repetitions each) using JDK 11 (default settings) on 8 cores of Xeon E5-2697A v4:
- the `split_scan` variant corresponds to commit d3a614a
- the `const_scan_lbq` variant corresponds to commit 18b4936 (scan requests queued using `LinkedBlockingQueue`)
- the `const_scan_stm` variant corresponds to commit 05ad765 (scan requests "queued" using STM transactions) 

plot-runs-samples
plot-runs-histograms
plot-runs-violins

### DIFF
--- a/benchmarks/scala-stm/src/main/scala/org/renaissance/scala/stm/Philosophers.scala
+++ b/benchmarks/scala-stm/src/main/scala/org/renaissance/scala/stm/Philosophers.scala
@@ -19,7 +19,13 @@ import org.renaissance.License
   defaultValue = "500000",
   summary = "Number of meals consumed by each philosopher thread"
 )
-@Configuration(name = "test", settings = Array("meal_count = 500"))
+@Parameter(
+  name = "block_meal_count",
+  defaultValue = "4096",
+  summary =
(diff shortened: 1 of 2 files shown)


lbulej:
Rebased to master and updated the commit ids in the description.

axel22 (approved):
LGTM.

PR #449, opened 2024-08-09
CI: fix use of continue-on-error and fail-fast

vhotspur:
According to documentation, continue-on-error [1] would mark every step as succesful inside the job, hence the job could never fail.

But we actually want fail-fast to be set to false, so that other jobs are not cancelled even when one job fails [2].

[1] https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error
[2] https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast


vhotspur:
Closing, superseded by #454

PR #450, opened 2024-08-21
Update SBT to version 1.9.9 and buildenv images and JDK versions in CI

lbulej:
Picks elements from #443 and adds a few related updates.

After some consideration, we actually drop the SBT binaries because the use case they were meant for (building Renaissance in a networkless environment)  does not seem to make much sense given the amount of dependencies that need to be downloaded. 

The CI was updated to use version 10 of `buildenv` images and the JDK versions tested were bumped to 22 (latest), { 11, 17, 21 } (legacy), and 23 (EA).


vhotspur (approved):
LGTM, thanks!

PR #452, opened 2024-09-09
CI: bump to v11 of Docker images

vhotspur:


lbulej (approved):
The checks passed, so LGTM.

PR #455, opened 2024-11-12
Use Spark to load CSV files and avoid extra IO

lbulej:
Use the Spark DataFrameReader to load the CSV files with movies and ratings to correctly load CSV files with quoted values. Specifically, `movies.csv` was loaded "by hand"  with splitting on coma, resulting in some movie titles being cut off and part of the movie title ending up in the "genres" field, but no harm was done, because the genres are not being used by the benchmark. Anyway, it also avoids the extra IO of copying the resource files to the scratch directory.

Given the above changes, the data initialization was made more localized to avoid keeping data that are not needed for benchmark execution after the initial setup.

Because the data are loaded using the `DataFrameReader` from Spark Session instead of the `textfile()` function from Spark Context (which uses `hadoopFile()` underneath), they are partitioned differently (there are more partitions), which seems to affect performance. To mimic the original behavior, we initially repartition the data using the same defaults as in `textfile()`.


farquet (approved):
LGTM!

PR #456, opened 2024-11-12
Update benchmark dependencies

lbulej:
Updates Scala, primary benchmark dependencies and a number of secondary dependencies. Partially builds on top of #443 (which it supersedes) and goes a bit further where possible. The `db-shootout` benchmark is now compatible with JDK upto 22 (interestingly, in standalone mode, it can actually run even on Java 24). This should bring us close to release.

Scala updates
- Scala 2.12 updated to `2.12.20`
- Scala 2.13 updated to `2.13.15`
- Scala 3 updated to `3.3.4`

Updates in `apache-spark`
- Spark updated from `3.5.0` to `3.5.3`
- Secondary dependencies
    - Arrow updated to `18.0.0`
    - Parquet updated to `1.14.3`
    - Netty updated to `4.1.114.Final`
    - Guava updated to `33.3.1-jre`

Updates in `db-shootout`
- `agrona` updated from `1.19.2` to `1.22.0` (later versions require Java 17)
- `mapdb` updated from `3.0.10` to `3.1.0`
- `h2-mvstore` updated from `2.2.224` to `2.3.232`
- `chronicle-map` updated from `3.22.9` to `3.26ea4` (supports Java 21+)
- Secondary dependencies
    - ASM updated to `9.7.1` (requires Java 11+)
    - `lz4` version `1.3.0` replaced with `lz4-java` version `1.8.0`
    - Guava updated to `33.3.1-jre`
    - JNA updated to `5.15.0`

Updates in `future-genetic`
- Jenetics updated from `5.2.0` to `6.3.0` (requires Java 11+)

Updates in `neo4j-analytics`
- `neo4j` updated from `5.12.0` to `5.25.1`
- Secondary dependencies
    - Arrow updated to `18.0.0`
    - ASM updated to `9.7.1` (requires Java 11+)
    - Caffeine updated to `3.1.8` (requires Java 11+)
    - gRPC updated to `1.68.1` (secondary)
    - JNA updated to `5.15.0`
    - Netty updated to `4.1.114.Final`
    - Netty Tomcat Native updated to `2.0.69.Final`
    - Parquet updated to `1.14.3`

Updates in `rx`
- `rxjava` updated from `3.1.8` to `3.1.9`

Updates in `scala-dotty`
- `scala3-compiler` updated from `3.3.1` to `3.3.4`

Updates in `twitter-finagle`
- Finagle updated from `22.12.0` to `24.2.0`
- Secondary dependencies
    - Caffeine updated to `3.1.8` (requires Java 11+)
    - Netty updated to `4.1.114.Final`

### DIFF
--- a/benchmarks/database/src/main/scala/org/renaissance/database/DbShootout.scala
+++ b/benchmarks/database/src/main/scala/org/renaissance/database/DbShootout.scala
@@ -14,7 +14,7 @@ import org.renaissance.License
 @Group("database")
 @Summary("Executes a shootout test using several in-memory databases.")
 @Licenses(Array(License.APACHE2))
-@SupportsJvm("18")
+@SupportsJvm("22") // Works on Java 24 in standalone mode.
 @Repetitions(16)
 @Parameter(name = "rw_entry_count", defaultValue = "500000")
 @Configuration(name = "test", settings = Array("rw_entry_count = 10000"))
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ The following is the complete list of benchmarks, separated into groups.
 
 - `db-shootout` - Executes a shootout test using several in-memory databases.
   \
-  Default repetitions: 16; APACHE2 license, MIT distribution; Supported JVM: 11 - 18
+  Default repetitions: 16; APACHE2 license, MIT distribution; Supported JVM: 11 - 22
 
 - `neo4j-analytics` - Executes Neo4j graph queries against a movie database.
   \
--- a/benchmarks/database/src/main/resources/simplelogger.properties
+++ b/benchmarks/database/src/main/resources/simplelogger.properties
@@ -1,1 +1,3 @@
 org.slf4j.simpleLogger.defaultLogLevel=warn
+# Silence warnings related to CPU cores with different number of hardware threads.
+org.slf4j.simpleLogger.log.net.openhft.affinity.impl.VanillaCpuLayout=error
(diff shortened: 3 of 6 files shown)


benchmarks/database/src/main/scala/org/renaissance/database/DbShootout.scala
```diff
@@ -14,7 +14,7 @@ import org.renaissance.License
 @Group("database")
 @Summary("Executes a shootout test using several in-memory databases.")
 @Licenses(Array(License.APACHE2))
-@SupportsJvm("18")
+@SupportsJvm("22") // Works on Java 24 in standalone mode.
```
farquet, line 17: in standalone mode, but not in regular mode?

farquet (approved):

benchmarks/database/src/main/scala/org/renaissance/database/DbShootout.scala
lbulej: Yes, in standalone mode, the `db-shootout` benchmark actually runs on Java 24, while with the harness it fails on Java 23. I suspect a classpath property issue when using a custom class loader, because Chronicle generates and compiles source code at runtime using the JDK compiler module and it fails with `ClassNotFound` exception for classes in jars that are definitely known to the custom class loader, but are not on the classpath. In standalone mode, all the jars are on the classpath, so Chronicle can find those it is looking for.

It's something to be investigated further, but I don't want it to become a blocker for the upcoming release.

lbulej: See #458

PR #460, opened 2024-11-19
Update Akka configuration to avoid registering JVM shutdown hooks and calling system exit on fatal error

lbulej:
Configures default logging levels, disables framework exit on fatal error (harness should do it) and disables JVM shutdown hooks (benchmark should handle it). The idea is that benchmark should not register global hooks that run outside the harness control. The benchmark appears to terminate the actor system properly, so there should be no need for Akka to do this.


vhotspur (approved):
LGTM.

Issue #457, opened 2024-11-13
gauss-mix randomly fails on CI


vhotspur:
Randomly but quite often (I would say one in five builds, perhaps) `gauss-mix` benchmark fails when run as part of the CI.

It does not fail with some assertion/exception but rather stays in some endless loop (probably) as the job time-outs without any apparent reason (just `gauss-mix` is the last benchmark that was started).

From my experience this happens more often (maybe only?) in the standalone runs, i.e. when the main (fat) JAR is unpacked and we run `single/gauss-mix.jar` directly.

I was unable to reproduce the failure locally (even using the build artifact). At the moment it happily runs for several days on my machine without a glitch.

I do not expect we would solve this any time soon but it should be known at least.

PR #462, opened 2024-11-20
CI: thread dump on benchmark timeout

vhotspur:
To debug #457 we could run each Java process in `timeout` that first sends `SIGQUIT` (`3`) which tells the JVM to dump all threads (thanks @ceresek for the suggestion) and then kills the process anyway.

This also shortens the execution time for failing jobs as `gauss-mix` is killed after 5 minutes instead of eating the maximum job time (30m).

This works on both Linux and Windows (I am not sure if the signal is somehow emulated as I have not seen it fail there yet but it does not fail anything on good runs), on MacOS the command is not installed but majority of the failures were on Linux anyway.

Already we have one instance of failure here: https://github.com/renaissance-benchmarks/renaissance/actions/runs/11932162022/job/33256738789?pr=462#step:9:397

### DIFF
--- a/tools/ci/bench-base.sh
+++ b/tools/ci/bench-base.sh
@@ -6,5 +6,5 @@ java -jar "$RENAISSANCE_JAR" --raw-list > list.txt
 
 for BENCH in $(cat list.txt); do
 	echo "====> $BENCH"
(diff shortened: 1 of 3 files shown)


lbulej (approved):
Looks like a good idea to me!

PR #463, opened 2024-11-21
Refactor parameter and variable names for clarity in the future-genetic benchmark

lovisek:
Renamed chromosome_count to population_size to better reflect its role. The benchmark generates a population of 50 genotypes, each containing a single chromosome with 200 double Genes. Evolution is computed for the population, and at the end, the best genotype is selected, and its chromosome is returned. Since the term chromosome_count=50 was misleading, this change improves clarity.

Renamed chromosomeRandom to genotypeRandom, as this random generator is responsible for creating new random genotypes, not chromosomes.


lbulej (approved):
LGTM. It changes an externally visible parameter, but I don't think users of Renaissance are commonly changing those. I think it should be enough to point out this change in the release notes.

PR #465, opened 2024-12-04
Close URLClassLoader instances to enable proper cleanup on NFS

### LINKED ISSUE #464: Harness leaves behind temporary files when run on NFS mount
Running Renaissance 0.16.0 with the current directory on an NFS mount leaves behind the `launcher` directory:

```
java -jar renaissance-gpl-0.16.0.jar --list
...
ls -R
.:
launcher-101515-11172672929640790059

./launcher-101515-11172672929640790059:
renaissance-harness_3

./launcher-101515-11172672929640790059/renaissance-harness_3:
lib

./launcher-101515-11172672929640790059/renaissance-harness_3/lib:
```

This is not a new issue, the same happened on 0.15.0.

lbulej:
This ensures that library JAR files stored in the launcher and harness scratch directories are closed before removing the scratch
directory roots. This was a problem on NFS, where the open files could not be unlinked from the directory (NFS kept around `.nfs*` temporaries), preventing the removal of the directories.

Also fixes the logging setup which used `StreamHandler` instead of `ConsoleHandler` for `stderr`, which resulted in closing of the `stderr` as a result of running the `java.util.logging` shutdown hook, effectively gagging `stderr`.

Closes #464

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/core/BenchmarkSuite.java
+++ b/renaissance-core/src/main/java/org/renaissance/core/BenchmarkSuite.java
@@ -283,6 +283,10 @@ public final class BenchmarkSuite {
       throw new ExtensionException("malformed URL(s) in classpath specification");
     }
 
+    //
+    // No need to explicitly close this URLClassLoader on exit, because it does not
+    // operate on files created by the harness that need to be deleted on exit.
+    //
     ClassLoader parent = ModuleLoader.class.getClassLoader();
     return new URLClassLoader(classPathUrls, parent);
   }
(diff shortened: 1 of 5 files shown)


ceresek:
I confirm this fixes #465 for me.

ceresek (approved):

vhotspur (approved):

lbulej:
Just for the sake of discussion, I considered an alternative where closing the `URLClassLoader` could be done explicitly through the `ExecutionDriver` after finishing execution of the benchmark.

On the one hand, this would close the files sooner and make the responsibility more explicit and localized, all of which is potentially a good thing. On the other hand, apart from the `ModuleLoader`, the upper layers don't really know that the class loader is actually a `URLClassLoader` (they would have to check if the class loader implements `Closeable`), so I kept it in the `ModuleLoader`.

Finally, we also create `URLClassLoader` instances for plugins, but don't keep track of those. This is because we don't manage the plugin files in any way (and are not responsible for deleting them). I added some comments along those lines.

PR #467, opened 2025-02-14
Restore logo in README

vhotspur:
Reverting the trigger happiness in #466 :-)


vhotspur:
> Sorry I did not catch that in https://github.com/renaissance-benchmarks/renaissance/pull/466, it wasn't immediately visible :-)

That was my fault, I have noticed it now mostly by accident. Not sure how should we prevent that, though.

PR #469, opened 2025-03-27
Add VM start in terms of nanoTime() to the CSV and JSON outputs

### LINKED ISSUE #468: Save `vmStartNanos` into (JSON) output
In #216 we have settled on an output that dumps `uptime_ns` in nanoseconds and includes `start_unix_ms` as a baseline so that `uptime_ns` can be aligned with external measurements. Some tools, such as `perf` (and therefore also `proftool`) use `CLOCK_MONOTONIC` (which is the value returned by `nanoTime` on Linux) directly, it would therefore be useful to also dump `vmStartNanos` from the code that computes the relationship between UNIX time and the reported `uptime_ns`. That way, `uptime_ns` could be correlated with external measurements based on `CLOCK_MONOTONIC` simply by adding the reported `vmStartNanos`.

lbulej:
The value is stored as `vm_start_ns` in the CSV output and `vm.start_ns` in the JSON output. It has no official semantics (unlike `vm_start_unix_ms`), but allows recovering the `nanoTime()` value associated with the `vm_start_unix_ms` value. It also bumps the JSON format version to 6.

BTW, only the last commit is the important one, the others are things that I had lying around for quite some time and thought to sneak them in.

Closes #468

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/BenchmarkParameter.java
+++ b/renaissance-core/src/main/java/org/renaissance/BenchmarkParameter.java
@@ -84,15 +84,15 @@ public interface BenchmarkParameter {
   List<Map<String, String>> toCsvRows();
 
   /**
-   * Interprets the value as a CSV-like table with rows separated by
-   * semicolons, columns separated by commas, and a header row with
-   * column names.
+   * Interprets the value as a CSV-like table and produces a {@link List}
+   * of objects obtained by applying a user-defined mapping function on
+   * each row. Each row is represented as a {@link Map} of column values
+   * (strings) keyed by column names. The CSV-like input has a header row
+   * with column names, the rows are separated by semicolons, and columns are
+   * separated by commas.
    *
-   * @param parser Function to parse a row from a {@link Map} containing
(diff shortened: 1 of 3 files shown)


ceresek (approved):
LGTM, thanks !

I kinda expected you to come up with some sort of `SystemInfo` container that would be used to ship information like `vmStartNano` to the result writers, or even be available as a singleton ... this looks ... just ... straightforward :-)

lbulej:
> I kinda expected you to come up with some sort of `SystemInfo` container that would be used to ship information like `vmStartNano` to the result writers, or even be available as a singleton ... this looks ... just ... straightforward :-)

I thought along those lines, but I realized that so far, we just need to pass a single "special" value (that the writers cannot easily get themselves if they are to have the same value).

The writers are independent of each other and with the harness being the sole user and owner of their interfaces, we decide how the interface looks like. If it turns out that we need to provide the writers with more (static) information like this, or if we create another writer that would duplicate the MBean queries the JSON writer does, we can easily change the interface to accept some sort of system info provider. Also, I understood that the sooner this is supported the better :-)

PR #470, opened 2025-04-04
CI: update images to v15 (JDK 25 EA)

vhotspur:


lbulej (approved):
LGTM. We seem to be fine on 24, I'll be checking 25 manually (EA is not subject to automatic CI runs).

Issue #474, opened 2025-07-10
Some benchmarks fail with jdk26 after JDK-8361426: (ref) Remove jdk.internal.ref.Cleaner


MBaesken:
9 benchmarks fail after the removal of Cleaner, for example als, chi-square.

als :
```
java.lang.ExceptionInInitializerError
	at org.apache.spark.unsafe.array.ByteArrayMethods.<clinit>(ByteArrayMethods.java:47)
	at org.apache.spark.memory.MemoryManager.defaultPageSizeBytes$lzycompute(MemoryManager.scala:261)
	at org.apache.spark.memory.MemoryManager.defaultPageSizeBytes(MemoryManager.scala:251)
	at org.apache.spark.memory.MemoryManager.$anonfun$pageSizeBytes$1(MemoryManager.scala:270)
	at scala.runtime.java8.JFunction0$mcJ$sp.apply(JFunction0$mcJ$sp.scala:17)
	at scala.Option.getOrElse(Option.scala:201)
	at org.apache.spark.memory.MemoryManager.<init>(MemoryManager.scala:270)
	at org.apache.spark.memory.UnifiedMemoryManager.<init>(UnifiedMemoryManager.scala:58)
	at org.apache.spark.memory.UnifiedMemoryManager$.apply(UnifiedMemoryManager.scala:207)
	at org.apache.spark.SparkEnv$.create(SparkEnv.scala:320)
	at org.apache.spark.SparkEnv$.createDriverEnv(SparkEnv.scala:194)
	at org.apache.spark.SparkContext.createSparkEnv(SparkContext.scala:284)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:478)
	at org.apache.spark.SparkContext$.getOrCreate(SparkContext.scala:2883)
	at org.apache.spark.sql.SparkSession$Builder.$anonfun$getOrCreate$2(SparkSession.scala:1099)
	at scala.Option.getOrElse(Option.scala:201)
	at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:1093)
	at org.renaissance.apache.spark.SparkUtil.setUpSparkContext(SparkUtil.scala:81)
	at org.renaissance.apache.spark.SparkUtil.setUpSparkContext$(SparkUtil.scala:54)
	at org.renaissance.apache.spark.Als.setUpSparkContext(Als.scala:56)
[...]

PR #475, opened 2025-07-16
Update CI images to v 16 (bump to Fedora 42)

vhotspur:
See https://github.com/renaissance-benchmarks/docker-buildenv/pull/32

### DIFF
--- a/.github/workflows/ea-jdk.yml
+++ b/.github/workflows/ea-jdk.yml
@@ -5,7 +5,7 @@ env:
 jobs:
   linux-ea:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


vhotspur:
@lbulej So this is consistently failing on `scala-stm-bench7` on JDK 11. I have tried to reproduce it locally and it fails with thread dump about every second run (interestingly enough, I have not noticed this with 0.16.0 against which I have tested the images before opening the buildenv PR).

lbulej:
Closing due to #474 moving to v17 images with the exception of OpenJDK 11 runs and plugin builds.

PR #481, opened 2025-09-25
Patch Chronicle Core, Chronicle Map, Spark Unsafe, and Hadoop Client API to improve compatibility with Java 26+

lbulej:
Patches Chronicle Core, Chronicle Map, Spark Unsafe to avoid reflective lookups for internal classes, such as `jdk.internal.ref.Cleaner` removed in Java 26-ea+6, which broke the `db-shootout` and Spark benchmarks on Java 26+.

To move away from patching byte code using ASM, the patching process uses tiny dedicated projects containing specific classes classes modified to remove the offending code. The previous patch to Hadoop Client API (#453) has been converted to use the same approach.


lbulej:
I want to test this on JDK26-ea, for which we need an updated buildenv images, but it seems to be going in the right direction.

lbulej:
> I want to test this on JDK26-ea, for which we need an updated buildenv images, but it seems to be going in the right direction.

I finally managed to massage the GHA configuration to my liking (one would not believe to what lengths we have to go to share a workflow-level setting between multiple job-level keys), we seem to be running not only on JDK 25, but also on [JDK 26-ea](https://github.com/renaissance-benchmarks/renaissance/actions/workflows/ea-jdk.yml).

farquet (approved):

PR #484, opened 2025-09-28
Trigger early loading of the UserGroupInformation.RealUser class to avoid crash on exit

lbulej:
One of Hadoop shutdown hooks apparently accesses the class, but at that time, the class loader providing access to the benchmark's dependency JARs is already closed and the files deleted, resulting in a class loading exception in the shutdown handler. Loading the class early ensures that the class will be available to the late shutdown hook.

If this turns out to be too fragile, there are some options concerning the Hadoop `ShutdownHookManager` class, but integrating them into Renaissance would be more difficult:
- Making the `executeShutdown()` method public to allow triggering the Hadoop shutdown hooks in Renaissance on benchmark shutdown. This will would also need making the `shutdownInProgress` field public so that we can prevent the system-wide shutdown hook from executing the Hadoop shutdown hooks again. This is also doable through reflection, but probably not preferred.
- Alternatively, the class could be modified to avoid registering a system wide shutdown hook in the class initializer and instead expose that hook for Renaissance to call directly on benchmark shutdown. 

Closes #471

Issue #471, opened 2025-04-16
Crash on exit in spark family of benchmarks


ceresek:
A fresh build of 487cb2072afb37661563e5102559303b5ba26b4c on OpenJDK 21 (Fedora 41) crashes on exit in Spark family of benchmarks:

```
> java -jar target/renaissance-gpl-0.16.0-17-g487cb20.jar --repetitions 3 --json json page-rank
NOTE: 'page-rank' benchmark uses Spark local executor with 12 (out of 24) threads.
====== page-rank (apache-spark) [default], iteration 0 started ======
GC before operation: completed in 18.343 ms, heap usage 708.966 MB -> 29.554 MB.
====== page-rank (apache-spark) [default], iteration 0 completed (4719.606 ms) ======
====== page-rank (apache-spark) [default], iteration 1 started ======
GC before operation: completed in 58.054 ms, heap usage 1.263 GB -> 149.017 MB.
====== page-rank (apache-spark) [default], iteration 1 completed (2496.722 ms) ======
====== page-rank (apache-spark) [default], iteration 2 started ======
GC before operation: completed in 40.076 ms, heap usage 524.396 MB -> 149.146 MB.
====== page-rank (apache-spark) [default], iteration 2 completed (2587.023 ms) ======
Exception in thread "Thread-3" java.lang.NoClassDefFoundError: org/apache/logging/log4j/message/SimpleMessage
	at org.apache.logging.log4j.message.AbstractMessageFactory.newMessage(AbstractMessageFactory.java:59)
	at org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:1980)
	at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:1946)
	at org.apache.logging.slf4j.Log4jLogger.warn(Log4jLogger.java:255)
	at org.apache.hadoop.util.ShutdownHookManager.executeShutdown(ShutdownHookManager.java:131)
	at org.apache.hadoop.util.ShutdownHookManager$1.run(ShutdownHookManager.java:95)
Caused by: java.lang.ClassNotFoundException: org.apache.logging.log4j.message.SimpleMessage
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:593)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	... 6 more
```

[...]

PR #485, opened 2025-10-14
Load dependencies directly from the main bundle instead of extracting them to scratch directory

lbulej:
This is an alternative to #484 to fix issues related to cleanup of the scratch directory on NFS (or on file systems, which do not cope well with deleting open files), i.e. #471 and #464, with the latter thought to have been fixed by #465.

Instead of extracting dependency JARs to a scratch directory and constructing `URLClassloader` for the extracted files, we just register a `URLStreamHandlerFactory` for URLs with a `resource` protocol and construct the `URLClassloader` using such URLs. The class loader then happily loads the JAR files directly from the main bundle. The underlying `ResourceURLConnection` caches the extracted JARs using weak references so that multiple concurrent URL connections can share the data while classes are being loaded.

This generally simplifies the `ModuleLoader`, which does not have to extract the JARs to the scratch directory and allows keeping the module class loader open. Because the class loader does not keep open files in the scratch directory by the time the harness exits, this avoids the problem with deleting open files.

The `db-shootout` and `scala-dotty` benchmarks needed minor modifications, because both the Java compiler (used by Chronicle in `db-shootout`) and the Dotty compiler (used in `scala-dotty`) expect actual files on the class path, not URLs. For these two benchmarks, we extract the dependency JARs to a library directory and construct the class path for them.

### DIFF
--- a/renaissance-core/src/main/java/org/renaissance/core/ResourceUrlConnection.java
+++ b/renaissance-core/src/main/java/org/renaissance/core/ResourceUrlConnection.java
@@ -1,23 +1,18 @@
 package org.renaissance.core;
 
-import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
 import java.net.URLStreamHandlerFactory;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 /**
- * An implementation of a {@link URLConnection} for resources. Serves primarily
- * to allow loading classes from resource JAR files. Because a single resource
- * may be accessed simultaneously from multiple connections, this class caches
- * the resource contents so that it can be shared between active connections.
- * The cached content is released after all connections using it are closed.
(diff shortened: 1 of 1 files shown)


lbulej:
Just to summarize our discussion outside GH, I'll split-off the part related to the directory cleanup (also reverting to keeping our class loaders open) into a separate PR to go into 0.16.1 so as to avoid class loading exceptions in shutdown hooks.

We'll keep the class-loading change for 0.17, possibly with some extra information on the maximum memory consumed during loading.

lbulej:
I have removed the caching part from the `ResourceURLConnection` implementation, because the extra complexity and initial memory consumption are likely not worth it. I have run the harness with `iotop` running on the side, and it behaves as expected &mdash; the physical I/O is eliminated by the operating system page cache, so the caching might have only helped to avoid logical I/O  (syscalls) and extra CPU work related "inflating" files from the JAR. Given that JAR files in the main bundle will be most likely stored (not deflated), it would only save the logical I/O.

Still, loading from the main bundle allows us to leave the class loaders open, we should not be deleting any JAR files that may be needed in shutdown hooks, and it does avoid write I/O due to the harness itself. Without the extra memory consumption (and given that helps alleviate the problem with annoying exceptions in shutdown hooks), we may actually consider including this in the 0.16.1 maintenance release.

lbulej:
BTW, I'll rebase this to master after #489 is merged to make the diff smaller.

PR #499, opened 2026-02-16
Document issues with finagle-chirper on large machines in README.md

simonis:


simonis:
Sure, and thanks for the patch @lbulej . I saw the CI failure but was a little confused and had not started to look into the root cause of it. Hope now everything will work.

PR #505, opened 2026-03-23
CI: bump to v19 (add 26 and 27-EA)

vhotspur:
This builds on latest JDK too (26) and adds 27-EA.


lbulej (approved):
Let's wait for the 27-EA run I triggered but otherwise LGTM. Thanks!

lbulej (approved):
Works like a charm now :-)

PR #472, opened 2025-04-30
Continue Ktor PR by @zavidnyi

nomisRev:
This PR continues the open PR from @zavidnyi

PR #506, opened 2026-04-02
Fix kotlin-ktor benchmark CI failures (see #472)

yalishevant:
## Summary

Fixes CI failures in the kotlin-ktor benchmark originally introduced in #472 by @zavidnyi and continued by @nomisRev.

The benchmark worked locally but failed on all CI platforms. Root causes:

- **Validation mismatch on odd CPU counts**: The validator computed expected task count using floating-point arithmetic, while `ClientManager.assignTasks` uses `.toInt()` truncation. On macOS CI runners (3 vCPUs), this produced a mismatch (expected 450 vs actual 400).
- **Fragile transitive dependencies**: `ktor-server-tests-jvm` and `ktor-server-auth-jvm` pulled in classes the benchmark actually needs (WebSockets, DefaultHeaders, client CIO) as transitive deps. Replaced with explicit dependencies.
- **DNS resolution in CI containers**: `localhost` can be unreliable in containerized environments. Switched to `127.0.0.1` in both server and client.
- **Race condition on server startup**: Clients could attempt to connect before the server was ready. Added a socket-based readiness check.
- **Potential indefinite hang**: `waitForMessage` had no timeout, risking CI job timeouts. Added a 30-second timeout via `withTimeout`.
- **Stale files**: Removed `output.csv` and `output.json` committed by mistake.


yalishevant:
Hi @vhotspur, could you please approve the workflow run for this PR (and for #507)?

This is a continuation of #472 — I've fixed the CI failures that @nomisRev encountered (validation mismatch on odd CPU counts, fragile transitive dependencies, server startup race condition). All checks pass locally, including base/standalone/JMH modes with different client counts.

yalishevant:
Just a gentle ping — all 21 CI checks are passing across the full matrix (as well as in #507). Would be great if you could take a look when you get a chance 🙂 

Happy to address any feedback!

lbulej:
> Just a gentle ping — all 21 CI checks are passing across the full matrix (as well as in #507). Would be great if you could take a look when you get a chance 🙂

Thanks for pushing this through! I'll have a look.

yalishevant:
Hi! Circling back on this one when you have a spare moment.

Quick status: the branch is still green across the full matrix (all 21 checks), rebased and ready — no changes needed on my end, and I'm happy to rebase again or address any review comments the moment you have feedback.

One bit of extra context on why we'd love to see this land: on the Kotlin side we work closely with several JDK/runtime vendors, and there's genuine interest from their performance teams in having Kotlin workloads represented in Renaissance. Ktor exercises coroutines, the Kotlin stdlib and a fairly realistic async server path - code shapes that differ from the existing Java workloads and give JIT/GC folks useful additional coverage. Getting this benchmark in-tree would let those teams track it as a first-class part of the suite rather than carrying it out-of-band.

Thanks again for offering to take a look, and let me know if there's anything I can do to make the review easier 🙂

lbulej:
> 
> One bit of extra context on why we'd love to see this land: on the Kotlin side we work closely with several JDK/runtime vendors, and there's genuine interest from their performance teams in having Kotlin workloads represented in Renaissance. Ktor exercises coroutines, the Kotlin stdlib and a fairly realistic async server path - code shapes that differ from the existing Java workloads and give JIT/GC folks useful additional coverage. Getting this benchmark in-tree would let those teams track it as a first-class part of the suite rather than carrying it out-of-band.
> 
> Thanks again for offering to take a look, and let me know if there's anything I can do to make the review easier 🙂

Apologies, I have been mostly occupied by shaping the work on completing the validation in existing benchmarks., but the interest in having more Kotlin-produced bytecode in Renaissance is shared. I hope to get things moving in that direction soon.

lbulej:
I went over the code and apart from the changes below, the impression from the benchmark is that it is currently a bit on the microbenchmark side: clients run a fixed set of independent, fire-and-forget tasks (join a chat and send a message, or create/reuse a direct chat and send a message) as fast as possible. It would be nice if clients had more varied behavior: listing and joining public chats, inviting contacts to group chats, posting messages in response to messages they receive, sending binary data to simulate pictures or videos, and so on.

That said, more varied client behavior would not necessarily exercise more of the framework, and without persistent storage, state would be mostly held in collections, unless we tried to use some in-memory database such as MapDb (which I believe is in Kotlin, but I'm not sure how muchin vogue it is).

Clients paced by incoming messages (instead of firing requests back to back) might stress the coroutine subsystem differently, and sending different kinds of data might throw serialization into the mix. Also, there is currently just one WebSocket endpoint, so there no routing complexity to speak of - is that something worth focusing on?

A richer workload would also need more thorough validation than counting successful tasks, for example hashing the contents of each chat, if a stable ordering can be established from clients reacting to messages they receive.

Nevertheless, there is an incentive to have a Kotlin-native workload, so unless I broke something with my changes, I would merge this as a staging benchmark. The benchmark would live in the active codebase, it would be runnable on demand, but it would be listed separately from the stable benchmarks, indicating that it may still change significantly.

Would you or somebody else in the Kotlin community be interested in pushing the benchmark further or identifying things in Kotlin runtime that would be nice to exercise within this benchmark?

## Fixes

- Task failures and cancellation were being caught and discarded (`catch (Throwable)`), both in client tasks and in the server's per-socket send helper. Failures now propagate instead of being swallowed.
- `waitForMessage`'s match result was discarded, so a task counted as successful even when the server's response did not match what was expected. The result now reaches `Client.run()` and gates the success count.
- A client used to invent a new random id per message (via `generateNonce()`) instead of keeping one identity, so the same client could appear under different names for group vs. direct messages. It also resent a rename command before every single message. Clients now share one `User` identity across all their tasks, set once via a prologue task.
- `User`'s display name was written under a lock on rename, but read elsewhere without any synchronization, so a rename was not guaranteed to become visible to other threads. Now backed by `AtomicReference`.
- A single, non-thread-safe `Random` was shared across tasks and clients running in different coroutines.
- `client_count = 1` crashed: direct-messaging excluded the sender from the recipient pool, leaving no one to pick when there was only one client. Self-messaging is now allowed.
- Removed a broken server-readiness polling loop: a `return@repeat` bug meant it never worked as intended, and it turned out to be unnecessary anyway, since `server.start()` already blocks until the server is ready.

## Workload

- Replaced `generateNonce()` (`SecureRandom`-backed, ignores `random_seed`) with the benchmark's own seeded random for both user ids and chat ids, so ids are reproducible across runs with the same seed.
- `ChatServer` now resets `chats`, `users`, and `userSockets` to empty maps on every repetition. `chats` is recreated from a fixed initial set of ids generated once from a seeded RNG, so repetitions start from the same state instead of inheriting a larger map capacity from a previous one.
- Random generators are now split off per client/task (`SplittableRandom`) instead of sharing or reseeding one instance, so each gets an independent, reproducible sequence.
- `ClientManager` is now the single source for expected task counts, instead of the same formula being duplicated in `KtorRenaissanceBenchmark`.
- `Client` now owns and closes its own `HttpClient`.

## Cleanups

- Split `setUpBeforeAll` into single-purpose steps.
- Dropped the redundant `user` prefix from `User.userId`/`User.userName` (now `id`/`name`).
- Renamed `private_message_fraction` to `direct_message_fraction` and updated the related code, matching the terminology already used throughout the codebase (`DirectMessageChat`, `DirectMessageClientTask`, etc.).
- Unified spelling (American) and merged/renamed a few files to match Kotlin's file-naming convention.
- Cleaned up `ChatApplication`'s comments: removed narrative comments and a couple that were factually wrong, dropped an unused logging config.
- Updated the README to match the current code and terminology.

## Other

- Removed unnecessary synchronization and generics: an `AtomicInteger` with no concurrent access to guard, `inline`/`reified` with no actual use.
- Removed `CreateChatCommand`/`AddUserToChatCommand`: neither was ever sent by any client task, and the server-side handler still used unseeded `generateNonce()`.
- Removed unused fields `Chat.description` and `ChatServer.directChatIdToCommonChatId`.

Issue #509, opened 2026-06-09
finagle-chirper times out on systems with high core number


MBaesken:
On a system with high core number (128) the finagle-chirper benchmark times out.  This happens on a Linux ppc64le Power11 system with 128 cores.  On other systems (also Linux ppc64le with less cores) the benchmark was successful.

Seems it cannot handle such high core numbers well .
The tests happened with OpenJDK 27/28 .

MBaesken:
[thread_dump_2.txt](https://github.com/user-attachments/files/28760269/thread_dump_2.txt)
[thread_dump_1.txt](https://github.com/user-attachments/files/28760270/thread_dump_1.txt)

MBaesken:
We had to increase the max heap (Xmx value) to 8G and also increase the limit of max. open files (ulimit -n).
Without the increased open file number limit we got a lot of exceptions like the one below.
They are gone now and the finagle chirper works.

```
Failure(Unable to create Channel from class class io.netty.channel.socket.nio.NioSocketChannel at remote address: localhost/127.0.0.1:36829. Remote Info: Not Available, flags=0x08) with RemoteInfo -> Upstream Address: Not Available, Upstream id: Not Available, Downstream Address: localhost/127.0.0.1:36829, Downstream label: :36829, Trace Id: 3d532e9d6f03ca25.0053ec3bb1855128<:e3bce6462d3da1f7 with Service -> :36829 with AppId -> mymachine
Caused by: com.twitter.finagle.ConnectionFailedException: Unable to create Channel from class class io.netty.channel.socket.nio.NioSocketChannel at remote address: localhost/127.0.0.1:36829. Remote Info: Not Available
        at com.twitter.finagle.netty4.ConnectionBuilder$$anon$1.operationComplete(ConnectionBuilder.scala:105)
        at com.twitter.finagle.netty4.ConnectionBuilder$$anon$1.operationComplete(ConnectionBuilder.scala:84)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:590)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:557)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:492)
        at io.netty.util.concurrent.DefaultPromise.addListener(DefaultPromise.java:185)
        at io.netty.channel.DefaultChannelPromise.addListener(DefaultChannelPromise.java:95)
[...]

PR #510, opened 2026-06-19
Update README after #509

vhotspur:
Just describe the issue with finagle-chirper timing-out as a known-issue in a more visible place.


lbulej (approved):
Looks good, thanks!

PR #512, opened 2026-06-22
Update scalafmt to version 3.8.3 to handle Scala 3 syntax

lbulej:
Using scalafmt 3.x is required to properly support Scala 3 sources.

The configuration attempts to reproduce scalafmt 2.6.4 output as closely as possible to minimize formatting changes. The changes in `gauss-mix`, `movie-lens`, and `fj-kmeans` are unavoidable but very minor.


vhotspur (approved):
Looks innocent enough :-)

PR #514, opened 2026-06-30
akka-uct: redesign workload and add advanced result validation

AdammBudai:
## Summary

The original `akka-uct` benchmark was a pure message-forwarding exercise that produced no numeric result, exercised no priority code path, and returned `Validators.dummy()`. This PR redesigns the benchmark from the ground up and adds advanced validation.

**Problems in the original:**
- Traversal workload produced no numeric result — nothing to validate
- Urgent traversal message was dead code: never sent by the root
- Priority mailbox (`ControlMessage`) never configured, so priority semantics were silently broken
- Substantial legacy code inherited from the Savina suite used atomic guards and countdown latches, introducing a race between latch completion and actor shutdown that led to a deadlock

**Redesigned workload:**
- Tree topology is pre-computed from a seeded RNG before any actors are spawned, giving a fully reproducible structure across runs
- Each repetition exercises `topology_count` (default 5) independent topologies with different seeds
- Actor system is created in `setUpBeforeEach` and destroyed in `tearDownAfterEach` so each iteration carries no residual state
- Tree construction follows a hierarchical message-based query protocol (`CheckExpansion` → `ExpansionQuery` → `CreateChildren`/`StayLeaf`); teardown mirrors it
- Each iteration performs exactly 3 scatter-gather reductions: 2 normal, 1 shock reduction that exercises the priority mechanism
- On the shock reduction, `UrgentUpdateTemperatureMessage` (a `ControlMessage`) is sent immediately after the normal measurement message; the priority mailbox reorders it ahead of the queue, propagating a shock temperature up through the tree
- Shock temperature is computed dynamically per topology so the validator interval is as tight as the topology permits
- Up to 10 urgent paths selected by a greedy minimax-divergence strategy (terminal leaf at ≥75% of max depth, paths spread across independent subtrees)

**Validation (advanced):**
- For each of the 5 topologies, two reference root temperatures are pre-computed analytically before any actor is spawned: nominal (no shock) and shocked (all urgent leaves at shock temperature)
- After each topology's actor run the actual root temperature must fall in the open interval (nominal − ε, shocked + ε), ε = 10⁻⁵
- Simultaneously verifies correctness of the temperature reduction across all nodes and effectiveness of priority message delivery
- On failure the validator reports the full expected range and the actual value

**Measurements:**

The speedup comes from pre-computing the topology (eliminating the per-iteration distributed construction round-trip to the root) and cleaner per-run actor lifecycle management.

akka_uct_distribution

akka_uct_histogram

akka_uct_median_per_run

akka_uct_variability


lbulej:
Please add some information on where the measurements were taken and what do you mean by the 10 steady-state iterations. Did you collect something like 50 and take last 10?

The speedup is not that interesting, because the amount of work has changed, but the spread of the values is significant. This may be the price for the reduction-style computations (tree construction, the temperature averaging, and tree destruction), because it requires context switches and somewhat more messages to travel.

On my ((2x10)+8)-logical-cpu laptop, this workload runs in < 5s with a peak CPU load ~20, which is why I was pushing for the reduced number of topologies to be exercised (originally there were 10, but non-deterministic even between benchmark repetitions).

BTW, we have had extensive back-and-forth about this code with Adam and I was pushing for the redesign. I have few more ideas for modifications, but that can wait after this is merged.

AdammBudai:
@lbulej I added info  about measurements

PR #515, opened 2026-06-30
db-shootout: add seeded workload and compound correctness validation

AdammBudai:
## Summary

cc @lbulej

The original `db-shootout` benchmark inherited a fixed workload from the `lmdbjava/benchmarks` project, returned `Validators.dummy()`, and had an MVStore file-growth bug that caused the database file to expand unboundedly across iterations. This PR replaces the workload with a seeded, delete-interleaved sequence and adds a compound validator covering all three database engines.

**Problems in the original:**
- MVStore (and MapDB in file-backed mode) used a copy-on-write format but the database was kept open across all repetitions, causing the on-disk file to grow by one full write set per iteration (~16× by the end of a default run)
- Fixed key array with no deletes: the surviving state after each iteration was always identical and trivially predictable

**Redesigned workload:**
- Workload data generated once in `setUpBeforeAll` using a fixed seed (42L) for full reproducibility
- Each thread owns an exclusive non-overlapping key slice; values are a per-key sequence of writes and deletes (`delete_probability` = 0.2 by default)
- Expected final state pre-computed by replaying the sequence: last non-delete value per key survives, deleted keys are absent
- Two reference values derived from the expected state: entry count and a polynomial rolling hash over sorted (key, value) pairs
- Database instances created in `setUpBeforeEach` and destroyed in `tearDownAfterEach`, eliminating the file-growth problem and ensuring a clean slate each iteration
- Raw `Thread` arrays replaced with `ExecutorService` + `CompletionService` for clean exception propagation
- Shared `DatabaseManager` base class gives all three engines a uniform lifecycle interface

**Validation (advanced):**
- After the write phase each engine's reader returns the database contents as a map
- Checksum recomputed from the actual map using the same polynomial hash
- Compound validator checks 6 conditions: entry count and checksum for each of MapDB, Chronicle Map, and MVStore independently
- Entry count mismatch catches missing or spurious keys; checksum mismatch catches wrong values under a correct key

|---|---|---|
| Median iteration time | 9 187 ms | 8 939 ms |
| Change | — | ~1.03× |

The validated workload does substantially more work per iteration than the original single bulk insert-and-read pass — a per-key write/delete sequence, a full-state read-back, and a compound checksum across all three engines — but the per-iteration data volume (`values_per_key`) was tuned down to compensate, bringing the median iteration time back in line with the original.
db_shootout_distribution
db_shootout_histogram
db_shootout_iterations_per_run
db_shootout_median_per_run


lbulej:
There seems to be a small merge conflict, can you fix it so that I can approve the CI run?

AdammBudai:
merge conflict solved

lbulej:
> merge conflict solved

Great! Now we need to scale the amount of work back to fit under 10 seconds.

AdammBudai:
@lbulej  amount of work is modified as requested

PR #516, opened 2026-06-30
finagle-chirper: add active user pool and end-to-end feed validation

AdammBudai:
## Summary

cc @lbulej

The original `finagle-chirper` benchmark bounded active concurrency by the CPU count regardless of the configured user population, and returned `Validators.dummy()` with no result checking. This PR introduces an explicit active user pool with deterministic state prediction and a fingerprint-based end-to-end validator.

**Problems in the original:**
- Number of active client threads equalled the number of available processors, making the user count parameter largely decorative: the vast majority of users were never active during a run
- A local integer digest was accumulated from response lengths but never compared against any reference

**Redesigned workload:**
- Explicit active user pool: a configurable number of users are drawn from the full population and designated active; only these users post messages during the timed iteration
- Active users distributed evenly across threads so each thread handles a group rather than a single user; thread count remains configurable (default: available processors)
- Decouples workload size from hardware thread count — a larger active pool increases the working set and contention on the shared feed map independently of the number of threads

**Deterministic state prediction:**
- Before any iteration the expected final feed state is pre-computed offline: inactive users retain their initial slice unchanged; for active users the same request-index arithmetic that governs the client loop is replayed single-threaded
- Predicted feeds for all users are reduced to a single fingerprint by sorting messages lexicographically within a concatenated representation and hashing — insensitive to insertion order, sensitive to the complete set of messages

**Validation (end-to-end):**
- After each iteration the actual feed contents are read from the master server and the same hash is computed
- A match confirms the server feed state is consistent with a correct serialised execution of the post sequence
- A mismatch indicates a lost, duplicated, or misattributed message under concurrent access to the shared feed map

|---|---|---|
| Median iteration time | 5 384 ms | 4 429 ms |
| Change | — | ~1.2× faster |

The validated workload initially ran substantially more total work per iteration than the original (each thread posts on behalf of a group of active users rather than a single one). `request_count` was tuned down to compensate, bringing the median iteration time back in line with — and now slightly under — the original, while keeping the active user pool and end-to-end validation intact.

finagle_chirper_distribution
finagle_chirper_histogram
finagle_chirper_median_per_run
finagle_chirper_variability

### DIFF
--- a/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
+++ b/benchmarks/twitter-finagle/src/main/scala/org/renaissance/twitter/finagle/FinagleChirper.scala
@@ -27,12 +27,13 @@ import java.net.InetSocketAddress
 import java.net.URLEncoder
 import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
 import java.util.Comparator
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.atomic.AtomicReference
 import scala.collection._
 import scala.collection.immutable.ArraySeq
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 import scala.collection.parallel.CollectionConverters._
 import scala.io.Source
 import scala.util.hashing.byteswap32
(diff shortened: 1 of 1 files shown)


lbulej:
@AdammBudai the test configurations may need tweaks to avoid tripping assertions, see the run failures

AdammBudai:
@lbulej fixed

lbulej:
I let the CI run on the fixed version, we'll probably need to tweak the workload size to push it back to around 5 seconds per repetition.

AdammBudai:
@lbulej  workload tweaked

PR #517, opened 2026-06-30
scala-stm-bench7: fix STMBench7 bugs and add invariant + opacity validation

AdammBudai:
## Summary

cc @lbulej

The original `scala-stm-bench7` benchmark invoked STMBench7 via its `main` method, mixing full data-structure initialisation into every measured iteration, and returned `Validators.dummy()`. The STMBench7 codebase contained code paths for structural invariant checking and sequential replay for opacity detection, but five independent bugs made them unreachable or incorrect. This PR fixes all five bugs, switches to a programmatic interface, and enables both validators.

**Problems in the original:**
- Each call to `run` re-invoked `main`, rebuilding the full assembly hierarchy from scratch and including initialisation cost in the measured time
- **Non-random operation selection:** the random call was commented out and replaced with a deterministic modulo formula, producing a fixed cycling pattern instead of the statistical distribution the spec requires
- **Crashing replay constructor:** `SequentialReplayThread` called a two-argument superclass constructor whose body was a single unconditional `throw` — every instantiation terminated before replay could begin
- **Non-cloneable random state:** the RNG extended `java.util.Random` whose seed is an `AtomicLong`; a shallow clone shared that field between original and copy, so failed transaction retries could not roll back their random consumption, causing the replay sequence to diverge
- **Missing parent identity check:** structural comparison confirmed a non-null parent reference but not that both sides carried the same identifier — two atoms could reference parents with different IDs and the check would pass
- **Swallowed validation exceptions:** the helper thread running invariant checks installed no uncaught exception handler; any `RuntimeError` was consumed by the JVM default handler, the calling thread observed a clean join and continued as if the check had passed

**Redesigned workload:**
- Data structure built once in `setUpBeforeAll` via a dedicated programmatic constructor
- Between iterations a `reset` method recreates thread instances and zeroes per-thread counters while leaving the data structure intact, removing initialisation cost from the timed path
- Random operation selection call restored
- `SequentialReplayThread` constructor fixed to call the correct four-argument superclass constructor
- RNG replaced with a custom linear congruential generator whose state is a primitive `long`; `clone()` produces a genuine independent copy
- Structural comparison extended to verify parent identifier equality
- Helper thread extended with an uncaught exception handler that rethrows any `RuntimeError` in the calling thread after join

**Validation (advanced):**
- **Invariant checking:** after each iteration a sequential traversal verifies structural consistency — every component referenced by an assembly is present, every connection between atomic parts is symmetric, all indices contain exactly the required elements
- **Opacity checking via sequential replay:** before each iteration the data structure is cloned; during the iteration each transactional operation records its identity, result, and an atomically assigned timestamp; after the iteration the per-thread logs are merged, sorted by timestamp, and replayed on the clone — any divergence indicates an opacity violation; the final clone state is also compared structurally against the main structure including parent identifier equality

|---|---|---|
| Median iteration time | 1 092 ms | 2 056 ms |
| Change | — | ~1.9× slower |

Removing data-structure reinitialisation from the timed path (previously repeated every iteration) made the per-iteration cost initially collapse to a small fraction of the original. However, opacity checking replays every logged operation single-threadedly and performs a full structural comparison after each iteration; that overhead scales with total operations across all threads rather than per-thread work, and does not shrink linearly with `operation_count`. `operation_count` was tuned back up to produce a measured workload comparable to the original, and is still being refined toward full parity.

scala_stm_bench7_distribution
scala_stm_bench7_histogram
scala_stm_bench7_median_per_run
scala_stm_bench7_variability

### DIFF
--- a/benchmarks/scala-stm/src/main/scala/org/renaissance/scala/stm/ScalaStmBench7.scala
+++ b/benchmarks/scala-stm/src/main/scala/org/renaissance/scala/stm/ScalaStmBench7.scala
@@ -5,15 +5,25 @@ import org.renaissance.Benchmark._
 import org.renaissance.BenchmarkContext
 import org.renaissance.BenchmarkResult
 import org.renaissance.BenchmarkResult.Validators
+import org.renaissance.BenchmarkResult.Assert
 import org.renaissance.License
 
 @Name("scala-stm-bench7")
 @Group("scala-stm")
-@Group("scala") // With Scala 3, the primary group goes last.
+@Group("scala")
 @Summary("Runs the stmbench7 benchmark using ScalaSTM.")
 @Licenses(Array(License.BSD3, License.GPL2))
 @Repetitions(60)
-@Parameter(name = "thread_count", defaultValue = "$cpu.count")
+@Parameter(
+  name = "thread_count",
+  defaultValue = "$cpu.count",
+  summary = "Number of worker threads to use"
+)
+@Parameter(
+  name = "operation_count",
+  defaultValue = "20",
+  summary = "Number of operations to perform"
+)
 @Configuration(name = "test")
 @Configuration(name = "jmh")
 final class ScalaStmBench7 extends Benchmark {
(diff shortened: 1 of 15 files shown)


lbulej:
We might want to tweak the workload to get a longer (similar to previous) running time.

AdammBudai:
@lbulej  workload tweaked, do we want come even close with iteration time to previous version?

lbulej:
> workload tweaked, do we want come even close with iteration time to previous version?

I think it's fine at this point.

BTW, do you have a hypothesis for the higher run-to-run variation? One of the changes was related to the randomness that generated the operations (which the previous version had as a simple modulo I think), so this workload has a broader set of operations being performed?

PR #520, opened 2026-09-18
Force log4j2 2.26.1 to fix security vulnerability in transitive deps

parttimenerd:
Both Apache Spark 3.5.3 and Neo4j 5.25.1 transitively pull in `log4j-core` 2.20.0, which is flagged by security scanners (fixed version: 2.25.3+). This is the same kind of fix as #367, but for the log4j2 (`org.apache.logging.log4j`) artifacts instead of legacy log4j 1.x.

The affected artifacts come from two sources:

- **Spark** (via `spark-network-common`): `log4j-api`, `log4j-core`, `log4j-1.2-api`, `log4j-slf4j2-impl`
- **Neo4j** (via `neo4j-logging`): `log4j-api`, `log4j-core`, `log4j-layout-template-json`

Both Spark 3.5.3 and Neo4j 5.25.1 hardcode `log4j.version=2.20.0` in their respective parent BOMs. This PR adds `dependencyOverrides` for all affected artifacts in the `apacheSparkBenchmarks` and `neo4jBenchmarks` SBT sections, forcing them to 2.26.1 (current latest stable).

A new release with this fix would be appreciated; the current `renaissance-0.16.1.jar` is being flagged on many machines.


lbulej (approved):
Thanks! I think we can do a maintenance (0.16.2) release before I start merging the bigger PRs that will impact the workloads more and that might still take a few weeks to settle.

Can you confirm that this prevents the jar bundle from being flagged?

parttimenerd:
I ask my colleague who originally found the issue (I'm just his trusty messenger in this case)

parttimenerd:
I used open-source tools to check the log4j version, and the tools didn't find anything.

parttimenerd:
But we're deploying it on our test servers today, so the security scanners should probably scan it in the next few weeks.

